PR #55, opened 2023-10-17
Move environment details to separate page

psoleckimoj:


server/routes/components.ts
```diff
@@ -20,33 +19,29 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
     return res.send(components)
   })
 
-  get('/queue/:componentId/*', async (req, res) => {
+  get('/queue/:componentId/:environmentName/*', async (req, res) => {
```
psoleckimoj, line 22: Take environment name as only ever getting one environment now

PR #57, opened 2023-10-19
HEAT-136: generate dependency dropdown

psoleckimoj:


server/@types/strapi-api.d.ts
```diff
@@ -1064,7 +1064,15 @@ export interface components {
       include_in_subject_access_requests: boolean
       github_project_teams_maintain?: unknown
       github_topics?: unknown
-      versions?: unknown
+      versions?: {
```
psoleckimoj, line 1067: Had to add this as not generated for some reason and caused TypeScript errors

PR #103, opened 2023-11-20
Fix types for veracode stuff

psoleckimoj:


server/routes/components.ts
```diff
@@ -19,6 +19,39 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
...
+      component.attributes.veracode_results_summary?.severity?.forEach(severity => {
+        severity.category.forEach(category => {
+          severityLevels[category.severity] += category.count
+        })
+      })
+
+      return {
+        name: component.attributes.name,
+        hasVeracode,
+        pass: component.attributes.veracode_policy_rules_status === 'Pass',
```
andrewrlee, line 46: is this one mapped from the compliance status?

andrewrlee (approved):
LGTM

server/routes/components.ts
psoleckimoj: Yep, I also discovered the way that types file gets generated means it stops at a particular level so I ended up having to amend it to give the entire severity bit as it just puts unknown :/

mattops: should be mapped from policy_rules_status

PR #106, opened 2023-11-22
Alee/drift radiator draft

andrewrlee:
This is a bit of a WiP

If there is no drift: 
<kbd>
Screenshot 2023-11-22 at 09 20 04
</kbd>

If there is a drift between environments:
<kbd>
Screenshot 2023-11-22 at 09 19 56
</kbd>

Still need to:
* Consider showing multiple radiators for service area / product / team
* Add colours
* Add to table/search?
* etc...

### DIFF
--- a/assets/js/driftRadiator.js
+++ b/assets/js/driftRadiator.js
@@ -6,6 +6,7 @@ const data = Object.fromEntries(environments.map(e => [`${e}`, undefined]))
 const updateVersion = ({ env, version, date, build, sha }) => {
   data[env] = { version, date, build, sha }
   $(`#${env}_build`).text(build)
+  $(`#${env}_date`).text(date)
 
   if (env != devEnvName && data[devEnvName]) {
     const devEnvSha = data[devEnvName].sha
@@ -16,16 +17,14 @@ const updateVersion = ({ env, version, date, build, sha }) => {
 
     const humanReadableDate = `${dayjs(devEnvDate).diff(thisEnvDate, 'day')} days behind dev`
 
-    $(`#${env}_date`).text(`${date} (${humanReadableDate})`)
-
     if (devEnvSha !== thisEnvSha) {
       $(`#${env}_details`).html(`
+      
+      <strong>Staleness:</strong> ${humanReadableDate}<br/>
       <strong>Differences:</strong>
       <a class="govuk-link--no-visited-state" href="https://github.com/ministryofjustice/${componentName}/compare/${thisEnvSha}...${devEnvSha}">View in GitHub</a>
     `)
     }
-  } else {
-    $(`#${env}_date`).text(date)
   }
 }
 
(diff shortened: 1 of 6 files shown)


server/views/pages/driftRadiator.njk
```diff
@@ -0,0 +1,40 @@
+{% extends "../partials/layout.njk" %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/panel/macro.njk" import govukPanel %}
+
+{% set pageTitle = applicationName + " - " + component.name %}
+{% set mainClasses = "app-container govuk-body" %}
+{% set devEnvName = component.environments[0].name %}
```
andrewrlee, line 7: assumes the first env in the list represents the head of development

server/views/pages/driftRadiator.njk
```diff
@@ -0,0 +1,40 @@
...
+{% block content %}
+
+{{ govukBackLink({
+  text: "Back",
+  href: "/components/" + component.name
+}) }}
+
+<h1 id="detailPageTitle">Deployment Drift Radiator</h1>  
+<h2>{{ component.name }}</h2>  
+    <p><a class="govuk-link--no-visited-state" href="https://app.circleci.com/pipelines/github/ministryofjustice/{{ component.name}}?branch=main">Circle CI pipeline</a></p>
```
andrewrlee, line 18: Assumes all services are on circle - I was hoping to link through to builds on each tile but as well as build number, also need workflow id, which isn't available

assets/js/driftRadiator.js
```diff
@@ -0,0 +1,77 @@
...
+
+  if (env != devEnvName && data[devEnvName]) {
+    const devEnvSha = data[devEnvName].sha
+    const thisEnvSha = data[env].sha
+
+    const devEnvDate = data[devEnvName].date
+    const thisEnvDate = data[env].date
+
+    const humanReadableDate = `${dayjs(devEnvDate).diff(thisEnvDate, 'day')} days behind dev`
+    const gitDiffUrl = `https://github.com/ministryofjustice/${componentName}/compare/${thisEnvSha}...${devEnvSha}`
```
andrewrlee, line 26: This allows you to see the difference in commits between this env and dev in github, e.g: [link](https://github.com/ministryofjustice/create-and-vary-a-licence-api/compare/f35bf67...fc8a524)

This provides one of the benefits of the deployment tool - allows you to work out what tickets are inflight.
(The other bit of the deployment tool is pulling out jira ticket ids and displaying their current ticket state)

server/views/pages/driftRadiator.njk
mattops: the old monitor called the circleci API to get the latest build status etc.

andrewrlee: it's a shame we don't have the pipeline id available as part of the build number (as well as the job number) 

Then we could link to this page:
https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-developer-portal/544

andrewrlee: there's [an api](https://circleci.com/docs/api/v2/index.html#operation/getJobDetails) that grabs the pipeline id for job.

You could then grab the pipeline and build up any url.

Which would support:
"As a user, I want to easily progress a known deployment from one environment to the next"

PR #124, opened 2024-02-05
HEAT-191: split filter dropdowns

psoleckimoj:


server/services/dataFilterService.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  async getTeamsDropDownList(teamName: string): Promise<MoJSelectDataItem[]> {
+    const strapiApiClient = this.strapiApiClientFactory('')
+    const teamsData = await strapiApiClient.getTeams()
+    const teams = teamsData.data.sort(sortData)
+    const teamsList = teams.map(team => {
+      return {
+        value: team.id.toString(),
+        text: team.attributes.name,
+        selected: formatMonitorName(team.attributes.name) === teamName,
```
andrewrlee, line 32: I wondered if we should have a better name for this - `toCode` or something. 

I guess ideally it would be derived in strapi so we can search by it too - rather than having to rely on numeric identifiers

andrewrlee (approved):
LGTM

PR #133, opened 2024-02-21
HEAT-179 Refactor active agencies logic

williamfalconeruk:
- moved logic to a utility function
- updated template to remove logic
- updated route to call format function

### DIFF
--- a/server/routes/components.ts
+++ b/server/routes/components.ts
@@ -90,10 +90,9 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
 
     const component = await serviceCatalogueService.getComponent(componentName)
     const environments = component.environments?.filter(environment => environment.name === environmentName)
-    let activeAgencies = ''
-    if (environments.length > 0) {
-      activeAgencies = formatActiveAgencies(environments[0].active_agencies as Array<string>)
-    }
+    const activeAgencies =
+      environments.length === 0 ? '' : formatActiveAgencies(environments[0].active_agencies as Array<string>)
+
     const displayComponent = {
       name: componentName,
       api: component.api,


server/routes/components.ts
```diff
@@ -90,11 +90,15 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
 
     const component = await serviceCatalogueService.getComponent(componentName)
     const environments = component.environments?.filter(environment => environment.name === environmentName)
-
+    let activeAgencies = ''
+    if (environments.length > 0) {
+      activeAgencies = formatActiveAgencies(environments[0].active_agencies as Array<string>)
+    }
```
psoleckimoj, line 96: Can make this a ternary, sorry it's fine as is, I just try to avoid simple `if` and `let` if I can 😄 

`const activeAgencies = environments.length === 0 ? '' : formatActiveAgencies(environments[0].active_agencies as Array<string>)`

server/utils/utils.ts
```diff
@@ -100,3 +100,16 @@ export const isValidDropDown = (req: Request, paramName: string): boolean => {
 
   return param !== undefined && /^[a-z0-9_-]+$/.test(param)
 }
+/**
+ * @param activeAgencies a list of agencies if available.
+ * @returns formatted version of the agencies.
+ */
+export const formatActiveAgencies = (activeAgencies: Array<string>) => {
```
psoleckimoj, line 107: Feel free to add a test or two separately.

psoleckimoj (approved):
LGTM!

server/routes/components.ts
williamfalconeruk: did comment then read back - 👍🏻

server/utils/utils.ts
williamfalconeruk: will do!

server/routes/components.ts
```diff
@@ -90,11 +90,14 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
 
     const component = await serviceCatalogueService.getComponent(componentName)
     const environments = component.environments?.filter(environment => environment.name === environmentName)
+    const activeAgencies =
+      environments.length === 0 ? '' : formatActiveAgencies(environments[0].active_agencies as Array<string>)
```
andrewrlee, line 95: super minor.. 

It seems like there is an assumption that there will always be a single environment on line 99. `environment: environments[0],`. So you could extract out an `environment` variable earlier and not have the extra `environments.length === 0` check. 

not sure the reason for the type assertion for `as Array<String>` but that could be that the type on  `formatActiveAgencies` doesn't support passing `undefined`, if so changing the parameter to `activeAgencies: Array<string> | undefined` might mean you don't need that.

andrewrlee (approved):
LGTM

server/routes/components.ts
williamfalconeruk: > It seems like there is an assumption that there will always be a single environment on line 99. environment: environments[0],. So you could extract out an environment variable earlier and not have the extra environments.length === 0 check.
I think there is an assumption that there will be at least one environment set up for a component, so I guess this may not be the case. I'll take a look in a future pass.

> not sure the reason for the type assertion for as Array<String> but that could be that the type on formatActiveAgencies doesn't support passing undefined, if so changing the parameter declaration on the function to activeAgencies: Array<string> | undefined might mean you don't need that.

ah yes, thanks for this, makes sense. The type returned from strapi is an undefined json object, so the above makes sense. Again i'll update in the next pass through.

PR #134, opened 2024-02-21
Move dropdowns to single function call in service

psoleckimoj:
Is this the sort of thing that you were thinking @andrewrlee ?


andrewrlee (approved):
LGTM

PR #145, opened 2024-03-15
HEAT-226 improve performance of monitor

andrewrlee:


server/routes/monitor.ts
```diff
@@ -151,17 +151,33 @@ export default function routes({ serviceCatalogueService, redisService, dataFilt
...
+      const lastMessageTime = relativeTimeFromNow(new Date(dateAdded))
+      const healthStatus = getHealthStatus(json)
 
-    const messages = await redisService.readStream(streams)
+      acc[key] = { version, lastMessageTime, healthStatus }
+      return acc
+    }, {})
 
-    return res.send(messages)
+    return res.send(JSON.stringify(Object.entries(result)))
```
andrewrlee, line 180: guess some of this could be pulled down into the service

andrewrlee: would make it a bigger change if we wanted to turn it back into  streams and similar to the amount of work the other endpoints are doing.

psoleckimoj (approved):
LGTM!

PR #149, opened 2024-04-09
HEAT-242: Veracode filters

psoleckimoj:


server/routes/components.ts
```diff
@@ -24,10 +25,47 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
...
+
+        if (passed && failed && !unknown && status !== null) {
+          return true
+        }
+
+        if (passed && !failed && unknown && (status === 'Pass' || status === null)) {
+          return true
+        }
+
+        if (!passed && failed && !unknown && status === 'Did Not Pass' && status !== null) {
```
andrewrlee, line 55: guess ` && status !== null` this is superfluous as you'd know it was `Did Not Pass`?

andrewrlee (approved):
LGTM

server/routes/components.ts
psoleckimoj: Good call, I'll look at doing a bit of remedial work next to tidy this up.

PR #160, opened 2024-05-23
New status history graph, with selectable range.

mattops:


assets/js/environment.js
```diff
@@ -12,35 +12,22 @@ let streamData = []
...
+  const traceUpX = []
+  const traceUpY = []
+  const traceUpText = []
+  const traceDownX = []
+  const traceDownY = []
+  const traceDownText = []
+  const traceUnknownX = []
+  const traceUnknownY = []
+  const traceUnknownText = []
+  var rangeStart
```
psoleckimoj, line 24: `let`

assets/js/environment.js
```diff
@@ -12,35 +12,22 @@ let streamData = []
...
+  for (let i = 0; i < stream.length; i++) {
```
psoleckimoj, line 25: A more functional way to do this would be `stream.forEach((streamItem, i) => {`

Then you have each item in the array without addressing by index/i.

assets/js/environment.js
```diff
@@ -53,39 +40,134 @@ function drawHealthChart(stream) {
...
+      traceUnknownText.push(healthOutput)
+    } else {
+      traceDownX.push(eventTimeStart)
+      traceDownY.push(eventStatus)
+      traceDownText.push(healthOutput)
     }
+  }
 
-    dataTable.addRows([['Status', stream[i].message.http_s, healthOutput, rowColour, eventTimeStart, eventTimeEnd]])
+  var traceUp = {
```
psoleckimoj, line 59: use `const` if the value won't change, `let` it will.

server/views/pages/environment.njk
```diff
@@ -140,6 +141,7 @@
   <script type="module" src="/assets/govukFrontendInit.js"></script>
   <script src="/assets/moj/all.js"></script>
   <script src="https://www.gstatic.com/charts/loader.js"></script>
+  <script src="https://cdn.plot.ly/plotly-2.32.0.min.js" crossorigin="anonymous" charset="utf-8"></script>
```
psoleckimoj, line 144: Better way is to add this to the CSP/helmet config 🙂

assets/js/environment.js
```diff
@@ -53,39 +40,134 @@ function drawHealthChart(stream) {
...
   }
-  google.visualization.events.addListener(chart, 'select', selectHandler)
 
-  chart.draw(dataTable, options)
+  Plotly.newPlot('healthChart', statusData, layout, { displayModeBar: false })
+  var myPlot = document.getElementById('healthChart')
+  myPlot.on('plotly_click', function (data) {
+    var healthText
+    for (var i = 0; i < data.points.length; i++) {
+      healthText = data.points[i].text
```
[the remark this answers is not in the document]
mattops: yes kind of, not 100% how it works but when you have a busy chart with lots of data, the area you click on might have more than one point so the `data` this function receives contains potentially more points.  In our case not so many - it seems to work anyway!

server/views/pages/environment.njk
mattops: How?

psoleckimoj (dismissed):
LGTM

psoleckimoj (approved):
LGTM!

PR #173, opened 2024-06-20
Fix for error when displaying IP allow lists containing custom IPs

mattops:


server/routes/components.ts
```diff
@@ -117,17 +117,23 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
 
     if (environments[0].ip_allow_list && environments[0].ip_allow_list_enabled) {
       const ipAllowListFiles = Object.keys(environments[0].ip_allow_list)
-      allowList.set('groups', [])
 
       ipAllowListFiles.forEach(fileName => {
         if (Object.keys(environments[0].ip_allow_list[fileName]).length > 0) {
```
psoleckimoj, line 122: Can probably ditch the `if` now as the `foreach` simply won't do anything if the length is 0. Or should behave that way

server/routes/components.ts
```diff
@@ -117,17 +117,23 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
...
+              const genericService = environments[0].ip_allow_list[fileName]['generic-service']
+              Object.keys(genericService).forEach(ipName => {
+                if (ipName !== 'groups') {
+                  allowList.set(ipName, genericService[ipName])
+                } else {
+                  allowList.set(ipName, Array.from([...new Set([...allowList.get(ipName), ...genericService[ipName]])]))
+                }
+              })
+            }
+            if (item !== 'generic-service') {
```
psoleckimoj, line 135: Just need an `else` rather than another `if`

mattops: not sure about that, what if we have both types?

mattops: ok I see

psoleckimoj (approved):
LGTM!

PR #178, opened 2024-07-01
HEAT-275 add service area diagram

andrewrlee:


assets/js/common.js
```diff
@@ -35,6 +35,7 @@ function createTable(id, ajaxUrl, orderColumn, orderType, columns) {
       [10, 25, 50, 75, 100, -1],
       [10, 25, 50, 75, 100, 'All'],
     ],
+    pageLength,
```
andrewrlee, line 38: allow overriding this so we can show all a larger page for service areas 
(There were only 12 so 2 hung off the next page)

server/views/pages/productDependencies.njk
```diff
@@ -5,6 +5,8 @@
 {% from "govuk/components/select/macro.njk" import govukSelect %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
+{% set containerClasses = "wide-screen" %}
```
andrewrlee, line 8: making the diagram appear the full width of the screen as they were a bit squidged

PR #185, opened 2024-07-18
Add commit messages to version history timeline

mattops:


assets/js/environment.js
```diff
@@ -215,12 +215,73 @@ function drawVersionChart(stream) {
...
+      const json = await response.json()
+      return json
+    } catch (error) {
+      console.error(error.message)
+    }
+  }
+
+  function getCommitMessages(json) {
+    const commitMessages = []
+    for (let i = 0; i < json.commits.length; i++) {
```
psoleckimoj, line 241: `json.commits.forEach((commit, i) => {`

Then rather than using `commits[i]` you can just use `commit` as each time round the loop it assigns the current array value to `commit` 👍

assets/js/environment.js
```diff
@@ -215,12 +215,73 @@ function drawVersionChart(stream) {
...
+
+      sha = json.commits[i]['sha'].substring(0, 6)
+      html_url = json.commits[i]['html_url']
+      commitMessages.push(`<a href=${html_url} target="_blank">${sha}</a> - ${message}`)
+    }
+    return commitMessages
+  }
+
   function selectHandler() {
     var selectedItem = versionChart.getSelection()[0]
```
psoleckimoj, line 258: `var` bad Matt. Either `let` or `const`

assets/js/environment.js
```diff
@@ -215,12 +215,73 @@ function drawVersionChart(stream) {
...
+
+      const response = await fetch(url, { headers: githubHeaders })
+      if (!response.ok) {
+        throw new Error(`Response status: ${response.status}`)
+      }
+
+      const json = await response.json()
+      return json
+    } catch (error) {
+      console.error(error.message)
```
psoleckimoj, line 235: To make it a consistent return, should probably return something `{}` ? when there is an error as you're not actually stopping the code on error here.

assets/js/environment.js
```diff
@@ -215,12 +215,73 @@ function drawVersionChart(stream) {
     },
   }
 
+  async function getData(componentName, previousCommitSha, commitSha) {
+    const url = `https://api.github.com/repos/ministryofjustice/${componentName}/compare/${previousCommitSha}...${commitSha}`
```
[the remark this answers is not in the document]
psoleckimoj: Same goes for other URLs

assets/js/environment.js
```diff
@@ -215,12 +215,73 @@ function drawVersionChart(stream) {
...
   function selectHandler() {
     var selectedItem = versionChart.getSelection()[0]
     if (selectedItem) {
+      var previousVersion = dataVersionTable.getValue(selectedItem.row - 1, 1)
       var version = dataVersionTable.getValue(selectedItem.row, 1)
       var time = dataVersionTable.getValue(selectedItem.row, 2)
-      $(`#versionOutputSelected`).text(`Version: ${version} - Deployed @ ${time}`)
+      var previousCommitSha = previousVersion.split('.').pop()
+      var commitSha = version.split('.').pop()
+      const promise = getData(componentName, previousCommitSha, commitSha)
```
psoleckimoj, line 265: No need to assign to a variable first, can just do `getDate(...).then....`

PR #193, opened 2024-08-02
First go at adding breadcrumbs

mattops:


psoleckimoj (approved):
LGTM!

My only suggestion is the data you pass into the templates. If you just use name or a couple of fields then pass those in rather than the entire product data 👍

PR #201, opened 2024-08-28
Excluding false positive from npm security alert for micromatch

andrewrlee:


psoleckimoj:
Not required anymore.

PR #207, opened 2024-09-11
Update dependencies 2024-09-11

psoleckimoj:
Previously there was an option in the config file for these suppressions. As I've mentioned several times, the way strapi outputs it's types leaves empty objects in places where there shouldn't be which is what causes this. Manually updating the type definition file is not an option because it'll be lost any time it's generated,


andrewrlee (approved):
LGTM

PR #225, opened 2024-09-25
Added filter for trivy page

Sandhya1874:


assets/js/trivy.js
```diff
@@ -1,54 +1,124 @@
...
+    })
+    if (!response.ok) {
+      throw new Error(`There was a problem calling: ${url}`)
+    }
+    return response.json()
+  }
+
+  fetchMessages = async componentNames => {
+    const vulnerabilities = await this.post('/reports/trivy/data', { componentNames })
+    // console.log(vulnerabilities)
```
andrewrlee, line 26: best to remove out commented out code

assets/js/trivy.js
```diff
@@ -1,54 +1,124 @@
...
+
+  start = async componentNames => {
+    await this.fetchMessages(componentNames)
+  }
+}
+
+jQuery(function () {
+  $('#updateProduct,#updateTeam,#updateServiceArea,#updateCustomComponentView').on('click', async e => {
+    // e.preventDefault(e)
+    // const productCode = document.getElementById('product').value
```
andrewrlee, line 44: think we probably should include the e.preventDefault but delete the other commented out code?

server/routes/reports.ts
```diff
@@ -68,8 +69,12 @@ export default function routes({ serviceCatalogueService }: Services): Router {
     return res.render('pages/trivy')
   })
 
-  get('/trivy/data', async (req, res) => {
+  post('/trivy/data', async (req, res) => {
+    const componentsToInclude = req.body.componentNames
+    // console.log(componentsToInclude)
```
andrewrlee, line 74: should really delete this

server/routes/reports.ts
```diff
@@ -68,8 +69,12 @@ export default function routes({ serviceCatalogueService }: Services): Router {
     return res.render('pages/trivy')
   })
 
-  get('/trivy/data', async (req, res) => {
+  post('/trivy/data', async (req, res) => {
```
andrewrlee, line 72: we should probably move this into the new `trivy` module?

server/routes/trivy.ts
```diff
@@ -0,0 +1,130 @@
...
+import { isValidDropDown } from '../utils/utils'
+
+export default function routes({ componentNameService, dataFilterService }: Services): Router {
+  const router = Router()
+
+  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+  // const post = (path: string, handler: RequestHandler) => router.post(path, asyncMiddleware(handler))
+
+  get('/', async (req, res) => {
+    if (req.query.updateServiceArea === '' && isValidDropDown(req, 'serviceArea')) {
```
andrewrlee, line 13: do we actually need this query parsing handling? 

I think e.preventDefault would mean these are never sent

server/views/pages/trivy.njk
```diff
@@ -39,6 +40,18 @@
...
   <script src="/assets/moj/all.js"></script>
+  <script src="/assets/mojFrontendInit.js"></script>
+  <script src="https://www.gstatic.com/charts/loader.js"></script>
+  <script src="/assets/js/dayjs/dayjs.min.js"></script>
+  <script src="/assets/js/dayjs/plugin/relativeTime.js"></script>
   <script src="/assets/js/common.js"></script>
   <script src="/assets/js/trivy.js"></script>
+<script nonce="{{ cspNonce }}">
+    jQuery(function () {
+      const components = {{components | toJson(2) | safe }}
```
andrewrlee, line 51: I was thinking.. another option would be to load up all the component data in `trivy.ts` and pass through here instead of just the names..

This would mean that we don't have the lag between loading the page and we could get rid of the /trivy/data endpoint and the API call which would simplify things quite a bit..

server/views/pages/trivy.njk
```diff
@@ -39,6 +40,18 @@
...
+      const csrfToken = "{{ csrfToken }}"
+      const renderer = new TrivyRenderer(csrfToken, 'list')
+      renderer.start(components)
+    })
+  </script>
 {% endblock %}
```
andrewrlee, line 57: the other thing worth mentioning is that's a lot of code without unit tests

assets/js/trivy.js
psoleckimoj: Yep

assets/js/common.js
```diff
@@ -25,7 +26,15 @@ function createTable({
 
     return `0000${major}`.slice(-4) + `0000${minor}`.slice(-4) + `0000${patch}`.slice(-4)
   }
-
+  const ajax = ajaxUrl && {
```
psoleckimoj, line 29: This will still error because you're passing it to the `extend` function on line 80 and will basically be passing it undefined if it doesn't exist. Ideally if there is no URL it should report an error and then return from the function.

assets/js/trivy.js
Sandhya1874: Removed the commented lines

server/routes/trivy.ts
psoleckimoj: preventDefault just stops the default action firing I believe https://api.jquery.com/event.preventDefault/

server/routes/reports.ts
psoleckimoj: Shouldn't this be moving into the trivy routes files?

assets/js/common.js
andrewrlee: it now uses either data or ajax so it works, there's no errors. I guess it would ideally throw an error if neither were supplied

server/routes/trivy.ts
Sandhya1874: Its sorted by https://github.com/ministryofjustice/hmpps-developer-portal/pull/227 which is merged in this PR .

assets/js/trivy.js
Sandhya1874: sorted with https://github.com/ministryofjustice/hmpps-developer-portal/pull/227

andrewrlee (approved):
LGTM

PR #226, opened 2024-09-25
HEAT-316 - add security scans to github actions

james-jdgtl:
Moving security scans from circle CI to github actions


andrewrlee (approved):
LGTM

PR #228, opened 2024-10-01
Added Veracode filtering part on screen

Sandhya1874:


assets/js/veracode.js
```diff
@@ -1,100 +1,155 @@
...
+    })
+  }
+
+  start = async componentNames => {
+    await this.fetchMessages(componentNames)
+  }
+}
+
+jQuery(function () {
+  $('#updateProduct,#updateTeam,#updateServiceArea,#updateCustomComponentView').on('click', async e => {
```
psoleckimoj, line 42: Doesn't need to be `async` as there's no `await` in the code

server/routes/reports.ts
```diff
@@ -15,7 +15,8 @@ export default function routes({ serviceCatalogueService }: Services): Router {
     return res.render('pages/veracode')
   })
 
-  get('/veracode/data', async (req, res) => {
+  post('/veracode/data', async (req, res) => {
+    const componentsToInclude = req.body.componentNames
     const resultFilters = getResultFilters(req.query.results as string)
```
psoleckimoj, line 20: This is a post method now, if it's the same form then they won't be in `req.query` anymore, rather `req.body` I think?

server/views/pages/veracode.njk
```diff
@@ -90,6 +100,16 @@
 {% block bodyEnd %}
   <script type="module" src="/assets/govukFrontendInit.js"></script>
   <script src="/assets/moj/all.js"></script>
+  <script src="/assets/mojFrontendInit.js"></script>
+  <script src="https://www.gstatic.com/charts/loader.js"></script>
```
psoleckimoj, line 104: What is using this?

server/routes/reports.ts
psoleckimoj: Also shouldn't this be moved into the veracode routes file?

server/routes/veracode.ts
```diff
@@ -0,0 +1,115 @@
+import { type RequestHandler, Router } from 'express'
+import asyncMiddleware from '../middleware/asyncMiddleware'
+import type { Services } from '../services'
+
+export default function routes({ componentNameService, dataFilterService }: Services): Router {
+  const router = Router()
+
+  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+
+  get('/', async (req, res) => {
```
psoleckimoj, line 10: You're not doing anything with the submitted values from the dropdowns here which is why they don't do anything. I think I did an include file or at least look at another file that uses them and it's routes files.

assets/js/veracode.js
```diff
@@ -1,100 +1,155 @@
...
+      body: JSON.stringify(body),
+    })
+    if (!response.ok) {
+      throw new Error(`There was a problem calling: ${url}`)
+    }
+    return response.json()
+  }
+
+  fetchMessages = async componentNames => {
+    const vulnerabilities = await this.post('/reports/veracode/data', { componentNames })
```
psoleckimoj, line 25: So moving this into `/veracode` route means updating this URL as well as moving the code otherwise it's pointing to the old route.

server/routes/reports.ts
psoleckimoj: Basically you have changed this to a POST from a GET but are looking in the query string for data but in [assets/js/veracode.js](https://github.com/ministryofjustice/hmpps-developer-portal/pull/228/files#diff-5c0294451916b91b895a702078c15639b36ea9b0d87cd4ce1078e366bd6aa80a) you're passing a different set of fields through the body as it's a POST.

PR #237, opened 2024-10-28
Heat 370 opensearch url

james-jdgtl:
Update the Kibana links within the environment page to OpenSearch

Validated locally - ready to be pushed to dev, I think.


andrewrlee (approved):
LGTM

PR #238, opened 2024-10-29
Removing veracode from reports.ts

Sandhya1874:


server/routes/veracode.ts
```diff
@@ -0,0 +1,86 @@
...
+import type { Services } from '../services'
+import { getFormattedName, veracodeFilters } from '../utils/utils'
+
+export default function routes({ serviceCatalogueService }: Services): Router {
+  const router = Router()
+
+  const get = (path: string, handler: RequestHandler) => router.get(path, asyncMiddleware(handler))
+  // const post = (path: string, handler: RequestHandler) => router.post(path, asyncMiddleware(handler))
+
+  get('/veracode', async (req, res) => {
```
psoleckimoj, line 13: So, you don't need `/veracode` here and all other routes in here, the initial `/veracode` is in `server/app.ts`. So basically `app.use('/veracode', veracodeRoutes(services))` is saying prefix every route in `veracodeRoutes` with `/veracode` which means what you've added here is `/veracode/veracode` 🙂 

Basically in this file remove `/veracode` in all the `get` and `post` statements you want to use 👍

server/routes/veracode.ts
```diff
@@ -0,0 +1,87 @@
...
+          report: hasVeracode ? component.attributes.veracode_results_url : 'N/A',
+          date: hasVeracode
+            ? dayjs(component.attributes.veracode_last_completed_scan_date).format('YYYY-MM-DD HH:mm')
+            : 'N/A',
+          codeScore: hasVeracode ? component.attributes.veracode_results_summary['static-analysis'].score : 0,
+          severityLevels,
+        }
+      })
+
+    return rows
```
psoleckimoj, line 64: Needs to be the same as the code you copied it from so `res.send(rows)` it's not a return 🙂

Sandhya1874: Thanks Paul , corrected it now

server/routes/veracode.ts
Sandhya1874: Corrected

PR #251, opened 2024-11-28
Updating types and adding monitoring library

andrewrlee:


server/config.ts
```diff
@@ -55,6 +55,7 @@ export default {
         response: Number(get('SERVICE_CATALOGUE_TIMEOUT_RESPONSE', 5000)),
         deadline: Number(get('SERVICE_CATALOGUE_TIMEOUT_DEADLINE', 5000)),
       },
+      healthPath: '/',
```
andrewrlee, line 58: there's no ping endpoint on strapi but this is what it uses as its k8s health probe

james-jdgtl (approved):
looks good to me!

PR #252, opened 2024-11-28
Added product id dropdown instead of product name and made write & br…

Sandhya1874:
…anch protection fields optoional in component Request form

### DIFF
--- a/server/utils/utils.ts
+++ b/server/utils/utils.ts
@@ -8,7 +8,7 @@ dayjs.extend(relativeTime.default)
 
 type HasName = { attributes?: { name: string } }
 type HasRepoName = { attributes?: { github_repo: string } }
-type HasproductId = { attributes?: { p_id: string } }
+type hasProductId = { attributes?: { p_id: string } }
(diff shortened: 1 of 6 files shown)


server/utils/utils.ts
```diff
@@ -99,6 +100,17 @@ export const sortComponentRequestData = (dataItem: HasRepoName, compareDataItem:
   return 0
 }
 
+export const sortProductIdData = (dataItem: hasProductId, compareDataItem: hasProductId) => {
+  if (dataItem.attributes.p_id < compareDataItem.attributes.p_id) {
+    return -1
+  }
+  if (dataItem.attributes.p_id > compareDataItem.attributes.p_id) {
+    return 1
+  }
```
andrewrlee, line 109: for comparing strings you can use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare).

andrewrlee (approved):
LGTM

PR #255, opened 2024-11-29
Improvements to the drift radiator

andrewrlee:
* Showing latest commit info alongside each commit

* Deriving drift and staleness from the latest commit rather than the dev version
    * This handles works better for components that use branch based deployments
    * Components within monorepos still use the dev version - as commits to the repo may mean no changes to an given service
    * Show diff based on latest main commit rather than dev version

* Flag services where we haven't received version updates for over a day
   * This is a bit of temporary thing whilst we fix the health ping

* Fix sorting so that situations where components where we can't calculate drift or staleness appear separate to those that are up to date

* Change the colour of services where we can't calculate drift/staleness to grey rather than green

* Update docs to include info about new calculation mechanism and behaviour around monorepos

* Removing a load of unused styling


assets/js/driftRadiator.js
```diff
@@ -38,14 +38,14 @@ class DeploymentRenderer {
...
-            environment.type !== 'dev' && Boolean(component.devEnvSha) && environment.sha !== component.devEnvSha
+          const gitDiffUrl = `https://github.com/ministryofjustice/${component.repo}/compare/${environment.sha}...${component.baseSha}`
+          const showDiff = Boolean(component.baseSha) && environment.sha !== component.baseSha
           const diffAnchor = showDiff ? `(<a class="govuk-link--no-visited-state" href="${gitDiffUrl}">diff</a>)` : ''
           return `
         <li>
           <a class="env govuk-link--no-visited-state" href="/components/${component.name}/environment/${environment.name}">
-              ${environment.name}
+              ${environment.name}${environment.daysSinceUpdated > 0 ? '&#129301' : ''}
+              <!-- ${environment.daysSinceUpdated} days since last update --!>
```
andrewrlee, line 48: this is shonky but will allow debugging issues without surfacing on the display

assets/js/driftRadiator.js
```diff
@@ -55,23 +55,36 @@ class DeploymentRenderer {
 
       $('#dependencyDriftRows').append(`<tr class= "govuk-table__row">
         <td class="govuk-table__cell"><a href="/components/${component.name}" class="govuk-link--no-visited-state">${component.name}</a></td>
-        <td class="govuk-table__cell govuk-table__cell--numeric" data-sort-value="${component.staleness.days}">
+        <td class="govuk-table__cell govuk-table__cell--numeric" data-sort-value="${component.staleness.sortValue}">
```
andrewrlee, line 58: allowing more complicated sorting

server/services/teamHealthService.ts
```diff
@@ -297,13 +304,23 @@ export default class TeamHealthService {
       .filter(env => env.type === 'prod')
       .sort((e1, e2) => e1.buildDate.getTime() - e2.buildDate.getTime())[0]
 
+    const commitDate = component.latest_commit?.date_time && startOfDay(new Date(component.latest_commit?.date_time))
+    const commitSha = component.latest_commit?.sha?.slice(0, 7)
+
+    const baseSha = commitSha
+    const baseDate = component.part_of_monorepo ? latestDevEnvironment?.buildDate || commitDate : commitDate
```
andrewrlee, line 311: if mono repo use the dev env like previously.. otherwise changing one component inside a mono repo will affect all components scores.

psoleckimoj (approved):
LGTM!

PR #258, opened 2024-12-05
Tidy up component request form HTML

psoleckimoj:


server/views/pages/componentRequestForm.njk
```diff
@@ -239,7 +240,7 @@
           classes: "govuk-fieldset__legend--s"
         },
         hint: {
-          text: "Must be @digital.justice.gov.uk"
+          text: "Must be an @digital.justice.gov.uk email account"
```
andrewrlee, line 243: `a`?

andrewrlee (approved):
LGTM

server/views/pages/componentRequestForm.njk
psoleckimoj: Then would need to say

Must an a Digital Justice email account, a at doesn't sound right?

PR #263, opened 2024-12-12
Components with Environment details page

Sandhya1874:


server/app.ts
```diff
@@ -16,6 +16,7 @@ import setUpValidationMiddleware from './middleware/setUpValidationMiddleware'
 import indexRoutes from './routes'
 import productRoutes from './routes/products'
 import componentRoutes from './routes/components'
+import componentEnvRoutes from './routes/componentsEnvDetails'
```
psoleckimoj, line 19: Personally I'd expand `env` or `Env` to environment so it is clear.

Sandhya1874: Updated

server/app.ts
```diff
@@ -16,7 +16,7 @@ import setUpValidationMiddleware from './middleware/setUpValidationMiddleware'
 import indexRoutes from './routes'
 import productRoutes from './routes/products'
 import componentRoutes from './routes/components'
-import componentEnvRoutes from './routes/componentsEnvDetails'
+import componentEnvironmentRoutes from './routes/componentsEnvDetails'
```
psoleckimoj, line 19: This was for everything where you're using `componentsEnv`, including the filenames etc. Just makes it clear.

Sandhya1874: Updated all file names and references to componentsEnvironment

psoleckimoj (approved):
LGTM!

assets/js/componentsEnvironmentDetails.js
```diff
@@ -0,0 +1,180 @@
...
+          healthPath: env.health_path === null ? 'Not set' : env.health_path,
+          appurl: env.url === null ? 'Not set' : env.url,
+          monitor: env.monitor === null ? 'Not set' : env.monitor.toString(),
+          active_agencies: env.active_agencies,
+          ip_allow_list_enabled: env.ip_allow_list_enabled === null ? 'Not set' : env.ip_allow_list_enabled.toString(),
+          include_in_subject_access_requests:
+            env.include_in_subject_access_requests === null
+              ? 'Not set'
+              : env.include_in_subject_access_requests.toString(),
+          modsecurity_enabled: env.modsecurity_enabled === null ? 'Not set' : env.modsecurity_enabled.toString(),
```
andrewrlee, line 20: do you need to explicitly call toString on these? I'd have thought they'd be ok

PR #265, opened 2024-12-16
Added Product and Teams details in componentsEnvironmentDetails page

Sandhya1874:


assets/js/componentsEnvironmentDetails.js
```diff
@@ -23,13 +24,31 @@ jQuery(function () {
           build_image_tag: env.build_image_tag,
           alert_severity_label: env.alert_severity_label === null ? 'Not set' : env.alert_severity_label,
           alerts_slack_channel: env.alerts_slack_channel === null ? 'Not set' : env.alerts_slack_channel,
+          product_id: item.attributes.product.data.attributes.p_id,
```
psoleckimoj, line 26: Personally I would pull this out so `const productData = item.attributes?.product?.data?.attributes` then you can refer to these as `product.name`, `product.slug` etc

Avoids a lot of duplication.

The `?` in the object reference will ensure `undefined` is returned rather than the code blowing up because that part of the object doesn't exist.

Sandhya1874: sorted

assets/js/componentsEnvironmentDetails.js
```diff
@@ -23,13 +26,28 @@ jQuery(function () {
...
           alerts_slack_channel: env.alerts_slack_channel === null ? 'Not set' : env.alerts_slack_channel,
+          product_id: productData ? productData.p_id : '',
+          product_name: productData ? productData.name : '',
+          product_slug: productData ? productData.slug : '',
+          team_name: teamData ? teamData.name : '',
+          team_slug: teamData ? teamData.slug : '',
         })
       })
     })
+    console.log(transformed)
```
psoleckimoj, line 37: Needs removing

Sandhya1874: Removed console.log

PR #290, opened 2025-02-18
Updates header and page width

simonnebMOJ:
Implements new [MOJ Header](https://design-patterns.service.justice.gov.uk/components/header/) and changes page width to match other HMPPS/DPS services.

Removed the (beta) text and added GOV.UK tag. Also removed Hmpps from app name as now longer needed


psoleckimoj (approved):
LGTM!

PR #294, opened 2025-02-27
Fix componentFormRequest email validation

Thomas-Geraghty:
Form description says that `@digital.justice.gov.uk` and `@justice.gov.uk` email's are valid, then complains saying you can only use a `@digital.justice.gov.uk` email. Considering these are going away, thought it was worth fixing!


Sandhya1874 (approved):
LGTM

PR #299, opened 2025-03-27
Move to try to fix types

andrewrlee:
Also adding new script to generate strapi types

This is a bit of WIP as the strapi def still breaks the app.. hoping to fix that in a follow up


server/data/strapiApiTypes.ts
```diff
@@ -1,52 +1,70 @@
...
+  data?: T[]
+  meta?: {
+    pagination?: {
+      page?: number
+      pageCount?: number
+      pageSize?: number
+      total?: number
+    }
+  }
+}
```
andrewrlee, line 13: Defined this generic type so that we can hopefully (🤞 ) tweak the `T` type without having to unwrap and wrap types.  

Maybe something like this
```ts
type HasTeam = { team?: Team }
export type Product = components['schemas']['Product'] & HasTeam
  ```

And we won't need to tweak all the calling sites

File: server/data/strapiApiTypes.ts [hunk 2]
```diff
@@ -1,52 +1,70 @@
...
 export type GithubProjectVisibility =
   components['schemas']['GithubRepoRequestRequest']['data']['github_project_visibility']
 
 export type GithubTeam = components['schemas']['GithubTeam']
 export type GithubTeamResponse = components['schemas']['GithubTeamResponse']
-export type GithubTeamListResponse = components['schemas']['GithubTeamListResponse']
 export type GithubTeamListResponseDataItem = components['schemas']['GithubTeamListResponseDataItem']
 export type GithubTeamRequest = components['schemas']['GithubTeamRequest']
+
+export type VeracodeResultsSummary = {
```
andrewrlee, line 58: Added type to represent the Veracode json blob without having to tweak the generated strapi types

server/data/strapiApiTypes.ts
andrewrlee: General idea being that we don't do any manual changing of the type file and instead do any patching to fix it here.. which should hopefully help us limp on until we migrate strapi versions.

I think in lots of places we should unwrap the page response in the strapi client and just return raw arrays - I don't think we really use the pagination in lots of places

james-jdgtl (approved):
Looks good!

PR #300, opened 2025-03-27
Bodge a fix for the latest strapi types

andrewrlee:
Now types can be regenerated using the `./generate-strapi-types.sh` script without the need for manually modifying the generated type defs.

This is quite hacky and will be difficult to maintain, the types still don't entirely reflect the data that's returned but we need to get access to the latest collections that have been added to the service catalogue. 

Still think it's worth investing in a bit of an anti corruption layer to decouple the rest of the application from the quirks of the strapi api and also prepare for the upgrade of strapi.. at this point we could remove strapi types entirely - if we had some nice conversion layer which was well tested with good example data.


server/data/strapiApiTypes.ts
```diff
@@ -11,14 +11,22 @@ export type ListResponse<T> = {
...
+type HasTeam = { team: TeamResponse }
+type HasServiceArea = { service_area: ServiceAreaResponse }
+type HasProduct = { attributes: { product?: { data: { attributes: ProductWithTeamAndServiceArea } } } }
 
-export type Product = components['schemas']['Product']
+type ProductWithTeamAndServiceArea = Omit<Omit<ProductResponse['data']['attributes'], 'team'>, 'serviceArea'> &
+  HasTeam &
+  HasServiceArea
+
+export type Product = Omit<components['schemas']['Product'], 'components'> & HasComponents
```
andrewrlee, line 23: This is all very mind melty but the idea was to patch the types rather than rely on manually editing the massive types file

server/routes/components.test.ts
```diff
@@ -129,7 +129,7 @@ const testComponent = {
       swagger_docs: '/swagger-ui.html',
     },
   ],
-} as Component
+} as unknown as Component
```
andrewrlee, line 132: This is a capitulation to the complexity 😓  - it would be possible to extend Component type to correctly include product info but it is a load of code and only used in this one place

server/routes/monitor.ts
```diff
@@ -15,6 +15,20 @@ type MonitorEnvironment = {
...
+    environments?: Environment[]
+    product?: {
+      data?: {
+        attributes?: {
+          p_id?: string
+        }
+      }
+    }
+  }
+}
```
andrewrlee, line 30: Added a little helper type here which is the minimum that is required for `getEnvironmentData` and a common subset of all the different types that were previously being passed in which means no longer need to cast

server/services/serviceCatalogueService.test.ts
```diff
@@ -193,7 +192,7 @@ describe('Strapi service', () => {
             attributes: { name: 'Component 2' },
           },
         ],
-      } as ListResponse<CustomComponentListResponseDataItem>
+      } as ListResponse<ComponentListResponseDataItem>
```
andrewrlee, line 195: these were a mistake from my last PR

server/services/teamHealthService.test.ts
```diff
@@ -268,7 +269,7 @@ describe('teamHealthService', () => {
             },
           },
         },
-      ])
+      ] as ComponentListResponseDataItem[])
```
andrewrlee, line 272: This is needed as the test data only specifies a subset of all fields - there's some technically required field that aren't actually relevant for this test

psoleckimoj:
@andrewrlee it's the issue with CMSes, they could do with a nice way of returning their data instead of just reflecting how they store it :( I wonder if there's a plugin 😀

psoleckimoj (approved):
LGTM!

PR #302, opened 2025-03-31
Starting to decouple app from strapi representation

andrewrlee:


server/utils/utils.ts
```diff
@@ -82,10 +85,6 @@ export const sortComponentRequestData = (dataItem: HasRepoName, compareDataItem:
   return dataItem.attributes.github_repo.localeCompare(compareDataItem.attributes.github_repo)
 }
 
-export const sortProductIdData = (dataItem: hasProductId, compareDataItem: hasProductId) => {
```
andrewrlee, line 85: this was identical to `sortData`

server/services/serviceCatalogueService.ts
```diff
@@ -145,7 +142,7 @@ export default class ServiceCatalogueService {
     serviceAreaId?: number
     serviceAreaSlug?: string
     withProducts?: boolean
-  }): Promise<ServiceArea> {
+  }): Promise<StrapiServiceArea> {
```
andrewrlee, line 145: This version of the strapi api has entirely different structure of responses depending on whether its a singular or multiple or embedded response. 

The next step would be to move this over to the same `ServiceArea` representation so we can have consistency and return the same type.

Dealing of which bits of a subtree are populated is a bit of a different kettle of fish - getting type safety over that might be challenging and not worth the complexity.

server/data/converters/serviceArea.test.ts
```diff
@@ -0,0 +1,74 @@
+import { ServiceAreaListResponseDataItem } from '../strapiApiTypes'
+import { convertServiceArea } from './serviceArea'
+
+test('check conversion', () => {
```
andrewrlee, line 4: This test demos the difference between the model response (that's we'd expose to the rest of the app) and the strapi response.

Also exposing a test factory data method for use elsewhere to reduce coupling of the data structure to the rest of the tests.

server/data/converters/modelTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export type ServiceArea = {
+  id: number
+  name: string
+  owner: string
+  serviceAreaId: string
+  slug: string
+  products: Product[]
+}
```
andrewrlee, line 26: A start of the simpler type structure

server/services/dataFilterService.ts
```diff
@@ -40,16 +40,14 @@ export default class DataFilterService {
...
     const serviceAreaList = serviceAreas.map(serviceArea => {
-      const formattedName = formatMonitorName(serviceArea.attributes.name)
+      const formattedName = formatMonitorName(serviceArea.name)
 
       return {
         value: useFormattedName ? formattedName : serviceArea.id.toString(),
-        text: serviceArea.attributes.name,
-        selected: useFormattedName
-          ? formattedName === serviceAreaName
-          : serviceArea.attributes.name === serviceAreaName,
```
andrewrlee, line 52: quite a trivial example of how it can simplify calling code

psoleckimoj (approved):
Nice! LGTM!

psoleckimoj (approved):
LGTM!

PR #305, opened 2025-04-22
feat/HEAT-537-HMPPS-Alert-Manager

eve-bramley:
New feature for viewing and filtering currently firing alerts


assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
+const applicationFilter = document.getElementById('application')
+const environmentFilter = document.getElementById('environment')
+const namespaceFilter = document.getElementById('namespace')
+const severityFilter = document.getElementById('severity')
+let previousDataJSON = ''
+
+jQuery(async function () {
+  //check url to see if any filters are currently applied
```
psoleckimoj, line 8: Normally linting would pick up needing a space after the `//` in comments 👍

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+}
+
+const updateAlerts = async () => {
+  try {
+    const currentData = await getAlerts()
+    //check against existing alerts
+    if (currentData !== previousDataJSON) {
+      alerts = currentData
+      previousDataJSON = currentData
+      let filters = getFiltersFromURL()
```
psoleckimoj, line 78: `const` as nothing is changing.

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+
+  setTimeout(watch, 5000)
+}
+
+const updateAlerts = async () => {
+  try {
+    const currentData = await getAlerts()
+    //check against existing alerts
+    if (currentData !== previousDataJSON) {
+      alerts = currentData
```
psoleckimoj, line 76: These should be passed in as args as we don't want the function changing data outside it's scope, same for `previousDataJSON`

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+}
+
+//add current filters to Url params
+function updateURLParams(filters) {
+  const params = new URLSearchParams()
+  if (filters.application.length) params.set('application', filters.application)
+  if (filters.environment) params.set('environment', filters.environment)
+  if (filters.namespace) params.set('namespace', filters.namespace)
+  if (filters.severity) params.set('severity', filters.severity)
+  history.replaceState(null, '', '?' + params.toString())
```
psoleckimoj, line 101: Better to use `?${params.toString()}`

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+  return {
+    application: params.get('application') || '',
+    environment: params.get('environment') || '',
+    namespace: params.get('namespace') || '',
+    severity: params.get('severity') || '',
+  }
+}
+
+//append tabledata to the #statusTable
+async function populateAlertTable(alerts) {
```
psoleckimoj, line 115: Does this need to be `async`?

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+  }
+}
+
+function getOptions(data, key) {
+  const set = new Set(data.map(a => a.labels[`${key}`]))
+  return Array.from(set).sort()
+}
+
+function removeOptions(selectElement) {
+  var i,
```
psoleckimoj, line 149: Either `let` or `const` not `var`

psoleckimoj: Also don't declare multiple variables on one line with a comma, not clear.

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+    L = selectElement.options.length - 1
+  for (i = L; i >= 0; i--) {
```
psoleckimoj, line 151: Would be better as something like:

```
selectElement.options.forEach((index, option) => selectElement.remove(index))
```

Rather than for loops, can't remember the syntax off the top of my head.

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+  const opt = document.createElement('option')
+  opt.value = ''
+  opt.textContent = ''
+  selectElement.appendChild(opt)
+}
+
+function renderDropdown(select, options, selectedValue, key) {
+  //clear dropDown options to prevent duplicates
+  removeOptions(select)
+  if (selectedValue != '') {
```
psoleckimoj, line 163: Always `!==` plus I don't think you need the if/else? The `if` you have inside the `if` part will work whether or not there is a `selectedValue`

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+    //append selected option
+    options.forEach(option => {
+      const opt = document.createElement('option')
+      opt.value = option
+      opt.textContent = option
+      if (option === selectedValue) opt.selected = true
```
psoleckimoj, line 169: Also try to make sure to always use `{}` the linter would usually pick up I think.

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+      opt.textContent = option
+      select.appendChild(opt)
+    })
+  }
+}
+
+function populateAlertFilters(dropDownFilters) {
+  dropDownFilters.forEach(filter => {
+    filter.forEach(item => {
+      let select = document.getElementById(`${item.type}`)
```
psoleckimoj, line 185: `const` as the value never changes.

assets/js/alerts.js
```diff
@@ -0,0 +1,228 @@
...
+  renderDropdown(applicationFilter, applications, currentFilters.application, 'application')
+  renderDropdown(environmentFilter, environments, currentFilters.environment, 'environment')
+  renderDropdown(namespaceFilter, namespaces, currentFilters.namespace, 'namespace')
+  renderDropdown(severityFilter, severities, currentFilters.severity, 'severity')
+}
+
+//filter alters by selected filters
+function applyFilters(alerts, filters) {
+  return alerts.filter(alert => {
+    if (filters.application && alert.labels.application !== filters.application) return false
```
psoleckimoj, line 207: Could just turn this into a big `return` statement and `||` the conditions of the multiple `if` statements.

psoleckimoj: The same was as in `isFiltersEmpty`

server/routes/alerts.ts
```diff
@@ -0,0 +1,70 @@
+import { type RequestHandler, Router } from 'express'
+import asyncMiddleware from '../middleware/asyncMiddleware'
+import type { Services } from '../services'
+import logger from '../../logger'
+import { getAlertName, getAlertType } from '../utils/utils'
+
+interface AlertListResponseDataItem {
```
psoleckimoj, line 7: Should really go in `/server/@types/index.d.ts` and imported so it can be used elsewhere if need be, or quite probably imported from the `strapi-api.d.ts` via `/server/data/strapiApiTypes.ts` ?

assets/js/alerts.js
eve-bramley: I went for the first option, do let me know if it's not ideal and I'll have a crack at the suggestion to remove/recreate the dropdown

assets/js/alerts.js
```diff
@@ -0,0 +1,210 @@
...
+    updateDropdowns(alerts, currentFilters)
+    updateAll(alerts, currentFilters)
+  })
+})
+
+// gets alerts every 5 seconds
+function updateAlerts(currentData, previousDataJSON) {
+  //check against existing alerts
+  if (currentData !== previousDataJSON) {
+    alerts = currentData
```
psoleckimoj, line 70: Still not happy with this, you're letting a function alert a global variable which is bad. Pass in `alerts` then update it in the correct scope as this function returns `currentData` anyways.

assets/js/alerts.js
```diff
@@ -0,0 +1,210 @@
...
+  opt.textContent = ''
+  selectElement.appendChild(opt)
+}
+
+function renderDropdown(select, options, selectedValue, key) {
+  //  clear dropDown options to prevent duplicates
+  removeOptions(select)
+  // append options
+  options.forEach(option => {
+    const opt = document.createElement('option')
```
psoleckimoj, line 154: `newOption`?

assets/js/alerts.js
```diff
@@ -0,0 +1,210 @@
...
+    opt.value = option
+    opt.textContent = option
+    if (option === selectedValue) {
```
psoleckimoj, line 157: Could just replace the `if` statement with `opt.selected = (option === selectedValue)`

server/routes/alerts.ts
```diff
@@ -0,0 +1,36 @@
...
+  get('/all', async (req, res) => {
+    const alerts = await getAlertsApi()
+    res.json(alerts)
+  })
+
+  return router
+}
+
+function getAlertsApi(): Promise<AlertListResponseDataItem[]> {
+  return fetch('http://localhost:8080/alertmanager/alerts?filter=businessUnit="hmpps"')
```
psoleckimoj, line 31: Shouldn't be hardcoding URLs as they can change per env, does it actually need the `http://localhost:8080` part?

assets/js/alerts.js
psoleckimoj: Also `alerts` looks to be a `const` so couldn't be updated this way?

eve-bramley:
refreshing the alert data was removing any yet-to-be applied selected values in the dropdown filters, so I've made a pendingValues object to save those unapplied values and moved all the dropdown state and behaviour inside the dropdownHandler object.

assets/js/alerts.js
```diff
@@ -0,0 +1,236 @@
...
+    const opt = document.createElement('option')
+    opt.value = ''
+    opt.textContent = ''
+    selectElement.appendChild(opt)
+  },
+  renderDropdown: function (select, options, selectedValue, key, isReset) {
+    if (isReset) {
+      this.pendingValues[key] = ''
+    }
+    selectedValue === select.value ? (this.pendingValues[key] = '') : (this.pendingValues[key] = select.value)
```
psoleckimoj, line 116: Won't stop a merge but might be more readable as:

```this.pendingValues[key] = selectedValue === select.value ? '' : select.value```

File: assets/js/alerts.js [hunk 18]
```diff
@@ -0,0 +1,236 @@
...
+    this.removeOptions(select)
+    // append options
+    options.forEach(option => {
+      const newOption = document.createElement('option')
+      newOption.value = option
+      newOption.textContent = option
+      selectedValue !== ''
+        ? (newOption.selected = option === selectedValue)
+        : (newOption.selected = option === this.pendingValues[key])
+      select.appendChild(newOption)
```
psoleckimoj, line 127: Again, it's more understandable that this is assigning something if you do the value conditionally

```
newOption.selected = selectedValue !== '' ? option === selectedValue : option === this.pendingValues[key]
```

psoleckimoj (approved):
Couple of changes ideally either in this PR or a quick tidy PR afterwards otherwise LGTM!

psoleckimoj (approved):
LGTM!

PR #310, opened 2025-04-28
Added alert to component page

cugorji-justice:
https://dsdmoj.atlassian.net/browse/HEAT-538

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -10,6 +10,7 @@ generic-service:
   env:
     INGRESS_URL: "https://developer-portal-dev.hmpps.service.justice.gov.uk"
     SERVICE_CATALOGUE_URL: "https://service-catalogue-dev.hmpps.service.justice.gov.uk"
+    ALERTMANAGER_URL: "https://alertmanager.live.cloud-platform.service.justice.gov.uk"
     NODE_ENV: "development"
 
   namespace_secrets:
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -10,7 +10,6 @@ generic-service:
   env:
     INGRESS_URL: "https://developer-portal-dev.hmpps.service.justice.gov.uk"
     SERVICE_CATALOGUE_URL: "https://service-catalogue-dev.hmpps.service.justice.gov.uk"
-    ALERTMANAGER_URL: "https://alertmanager.live.cloud-platform.service.justice.gov.uk/api/v2"
     NODE_ENV: "development"
 
   namespace_secrets:
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -10,7 +10,7 @@ generic-service:
   env:
     INGRESS_URL: "https://developer-portal-dev.hmpps.service.justice.gov.uk"
     SERVICE_CATALOGUE_URL: "https://service-catalogue-dev.hmpps.service.justice.gov.uk"
-    ALERTMANAGER_URL: "https://alertmanager.live.cloud-platform.service.justice.gov.uk"
+    ALERTMANAGER_URL: "https://alertmanager.live.cloud-platform.service.justice.gov.uk/api/v2"
     NODE_ENV: "development"
 
   namespace_secrets:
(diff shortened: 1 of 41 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -10,6 +10,7 @@ generic-service:
   env:
     INGRESS_URL: "https://developer-portal-dev.hmpps.service.justice.gov.uk"
     SERVICE_CATALOGUE_URL: "https://service-catalogue-dev.hmpps.service.justice.gov.uk"
+    ALERTMANAGER_URL: "https://alertmanager.live.cloud-platform.service.justice.gov.uk/api/v2"
```
psoleckimoj, line 13: If this is the same in all environments you can just put it in `values.yaml` 👍

cugorji-justice: I dont know if its same but I will confirm and fix

server/routes/components.ts
```diff
@@ -50,8 +63,31 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
       alerts_slack_channel: alertsSlackChannel,
       github_enforce_admins_enabled: component.github_enforce_admins_enabled,
       standardsCompliance: component.standards_compliance,
+      alerts: [] as DisplayAlert[],
     }
 
+    let alerts: DisplayAlert[] = []
+    const alertManagerUrl = process.env.ALERTMANAGER_URL
```
psoleckimoj, line 70: You'd pull the alert manager URL into `/server/config.ts` like all the other env vars then in here you'd just include config and pull it out from that 👍

server/routes/components.ts
```diff
@@ -50,8 +63,31 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
...
+    if (alertManagerUrl) {
```
psoleckimoj, line 71: Also I'd not bother with this, it's in config so should always be there and you have a `try/catch` block anyway.

server/views/pages/component.njk
```diff
@@ -23,6 +23,71 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 data-test="detail-page-title">{{ component.name }}</h1>
+    {% if component.alerts and component.alerts.length %}
```
psoleckimoj, line 26: You can do `{% if component.alerts | length %}`

server/views/pages/component.njk
```diff
@@ -23,6 +23,71 @@
...
+                      <tr class="govuk-table__row">
+                        <th class="govuk-table__header">Alert Name</th>
+                        <th class="govuk-table__header">Environment</th>
+                        <th class="govuk-table__header">Summary</th>
+                        <th class="govuk-table__header">Message</th>
+                      </tr>
+                    </thead>
+                    <tbody class="govuk-table__body">
+                      {% for alert in component.alerts %}
+                        {% if not loop.first %}
```
psoleckimoj, line 67: Can I ask what is in the first alert that doesn't want showing here?  Having a table with one entry is fine and would simplify this template no end.

cugorji-justice: there was a scenario where the alerts were a lot so the idea was just to show the first record on the table. so when the show more drop down is clicked, no need to repeat the first record again, it just needs to show the rest.

server/views/pages/component.njk
```diff
@@ -23,6 +23,71 @@
...
+          <thead class="govuk-table__head">
+            <tr class="govuk-table__row">
+              <th class="govuk-table__header">Alert Name</th>
+              <th class="govuk-table__header">Environment</th>
+              <th class="govuk-table__header">Summary</th>
+              <th class="govuk-table__header">Message</th>
+            </tr>
+          </thead>
+          <tbody class="govuk-table__body">
+            {% set firstAlert = component.alerts[0] %}
```
psoleckimoj, line 39: Still confused about the logic around different displays if there are 1 or 2+ alerts?

mattops (approved):
LGTM

server/routes/components.ts
```diff
@@ -56,8 +70,28 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
...
     }
 
+    let alerts: DisplayAlert[] = []
+    try {
+      const url = `${config.apis.alertManager.url}?filter=application="${componentName}"`
+      const resp = await fetch(url)
+      const parsed = await resp.json()
+      const dataArray = Array.isArray(parsed) ? (parsed as Alert[]) : []
+      alerts = dataArray
+        .filter(a => a.status.state === 'active')
```
psoleckimoj, line 83: `a` should be a meaningful name

psoleckimoj (approved):
LGTM!

PR #317, opened 2025-05-12
🩹  Temp disable alertmanager on components page- the AM endpoint is broken and the dev portal does not fail gracefully. 

mattops:


psoleckimoj (approved):
LGTM!

PR #321, opened 2025-05-15
Feat/HEAT-537-HMPPS-Alert-Manager routes refactor

eve-bramley:


server/data/alertsApiClient.ts
```diff
@@ -0,0 +1,14 @@
+import config from '../config'
+import { AlertListResponseDataItem } from '../@types'
+
+export default class AlertsApiClient {
+  async getAlerts(): Promise<AlertListResponseDataItem[]> {
+    const alertManagerEndpoint = `${config.apis.alertManager.url}/alerts`
+    const urlFilter = 'filter=businessUnit="hmpps"'
+    return fetch(`${alertManagerEndpoint}?${urlFilter}`)
```
psoleckimoj, line 8: We don't use `fetch`, have a nosey at the StrapiApiClient, we use a standard `RestClient` and you should just be able to copy that for the most part 🙂

psoleckimoj: Incidentally `async` `await` is preferable to `.then()` etc. Easier to understand and follow and can be wrapped in `try/catch`

server/data/index.ts
```diff
@@ -11,14 +11,16 @@ initialiseAppInsights()
...
 
 import StrapiApiClient from './strapiApiClient'
+import AlertsApiClient from './alertsApiClient'
 
 type RestClientBuilder<T> = (token: string) => T
 
 export const dataAccess = () => ({
   applicationInfo,
   strapiApiClientBuilder: (() => new StrapiApiClient()) as RestClientBuilder<StrapiApiClient>,
+  alertsApiClient: new AlertsApiClient(),
```
psoleckimoj, line 21: This would then follow the pattern on line 20 🙂

server/services/alertsService.ts
```diff
@@ -0,0 +1,12 @@
+import type { AlertsApiClient, RestClientBuilder } from '../data'
+import { AlertListResponseDataItem } from '../@types'
+
+export default class AlertsService {
+  constructor(private readonly alertsApiClient: RestClientBuilder<AlertsApiClient>) {}
+
+  async getAlerts(): Promise<AlertListResponseDataItem[]> {
+    const alertsApiClient = this.alertsApiClient('')
```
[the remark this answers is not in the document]
eve-bramley: I get an error: An argument for 'token' was not provided.

psoleckimoj (approved):
LGTM!

PR #323, opened 2025-05-21
fixed component not loading with error

cugorji-justice:
https://dsdmoj.atlassian.net/browse/HEAT-629

### DIFF
--- a/helm_deploy/hmpps-developer-portal/values.yaml
+++ b/helm_deploy/hmpps-developer-portal/values.yaml
@@ -29,8 +29,6 @@ generic-service:
     REDIS_TLS_ENABLED: 'true'
     APPLICATIONINSIGHTS_CONNECTION_STRING: 'InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/'
     ALERTMANAGER_URL: "http://monitoring-alerts-service.cloud-platform-monitoring-alerts:8080/alertmanager"
-    SERVICE_CATALOGUE_TIMEOUT_RESPONSE: 5000
-    SERVICE_CATALOGUE_TIMEOUT_DEADLINE: 5000
 
   # Pre-existing kubernetes secrets to load as environment variables in the deployment.
   # namespace_secrets:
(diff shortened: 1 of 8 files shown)


helm_deploy/hmpps-developer-portal/values.yaml
```diff
@@ -29,7 +29,9 @@ generic-service:
     REDIS_TLS_ENABLED: 'true'
     APPLICATIONINSIGHTS_CONNECTION_STRING: 'InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/'
     ALERTMANAGER_URL: "http://monitoring-alerts-service.cloud-platform-monitoring-alerts:8080/alertmanager"
-    
+    SERVICE_CATALOGUE_TIMEOUT_RESPONSE: 50000
+    SERVICE_CATALOGUE_TIMEOUT_DEADLINE: 50000
```
psoleckimoj, line 33: 50 seconds is a long timeout? Also you want to set it https://github.com/ministryofjustice/hmpps-developer-portal/blob/74e1cb933aa2dbf9a5186cdee53fdfe36feb2037/server/config.ts#L55 as a default value if you're changing it everywhere?

server/routes/components.ts
```diff
@@ -75,19 +68,14 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
...
-        .filter(alert => alert.status.state === 'active')
+      const allAlerts = await alertsService.getAlertsForComponent(componentName)
+      alerts = allAlerts
+        .filter(alert => alert.status?.state === 'active')
         .map(alert => ({
-          alertname: alert.labels.alertname,
-          environment: alert.labels.environment,
-          summary: alert.annotations.summary,
-          message: alert.annotations.message,
+          alertname: alert.labels!.alertname!,
```
psoleckimoj, line 75: Could you not use `alert.labels?.alertName ?? ''` which provides a default when there is a null or undefined value?

helm_deploy/hmpps-developer-portal/values.yaml
cugorji-justice: So I observed that it when when i load the component page it times out often because the response take long to return. So i increased the time out gradually until I had little to no timeout at 50sec.

So what would you advise? Revert?

server/routes/components.ts
cugorji-justice: Yeah, i could, I will change this now

helm_deploy/hmpps-developer-portal/values.yaml
```diff
@@ -29,7 +29,9 @@ generic-service:
     REDIS_TLS_ENABLED: 'true'
     APPLICATIONINSIGHTS_CONNECTION_STRING: 'InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY);IngestionEndpoint=https://northeurope-0.in.applicationinsights.azure.com/;LiveEndpoint=https://northeurope.livediagnostics.monitor.azure.com/'
     ALERTMANAGER_URL: "http://monitoring-alerts-service.cloud-platform-monitoring-alerts:8080/alertmanager"
-    
+    SERVICE_CATALOGUE_TIMEOUT_RESPONSE: 5000
```
psoleckimoj, line 32: These are already set as defaults in `config.ts` I believe?

psoleckimoj (approved):
LGTM!

eve-bramley (approved):
LGTM

PR #326, opened 2025-06-04
feat/HEAT-609-Alert-Manager-Environments

eve-bramley:
Consolidate similar environments for consistency with the rest of the dev portal
https://dsdmoj.atlassian.net/browse/HEAT-609

### DIFF
--- a/server/routes/alerts.ts
+++ b/server/routes/alerts.ts
@@ -15,12 +15,12 @@ export default function routes({ alertsService }: Services): Router {
     // Get alerts to determine which environments actually have data
     const alertsData = await alertsService.getAlerts()
 
-    // Extract unique canonical environment values from actual alerts
+    // Extract unique canonical environment values from actual alerts and deduplicate
     const alertEnvironments = [
       ...new Set(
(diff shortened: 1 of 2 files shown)


assets/js/alerts.js
```diff
@@ -128,6 +125,32 @@ const dropdownHandler = {
...
+    // Only update selection without rebuilding the dropdown
+    if (isReset) {
+      select.selectedIndex = 0
+      this.pendingValues[key] = ''
+      return
+    }
+
+    if (selectedValue) {
+      // Find and select the option that matches the selectedValue
+      for (let i = 0; i < select.options.length; i++) {
```
psoleckimoj, line 138: Don't use `for` loops where possible, should be able to use `select.options.forEach()`, then don't use the `break`, just `return`, or might even be able to use `select.options.findIndex()` possibly. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

eve-bramley: Function wasn't in use so has been removed

psoleckimoj (approved):
LGTM!

PR #327, opened 2025-06-04
Fixing issue where we occasionally show stale commit data in the drift view

andrewrlee:
We were showing stale data, if there has been a commit since the last time the github discovery script runs

Previously we preferred the main commit as it works better for repos that have no consistent dev environment.

Added a few more test cases too


server/services/teamHealthService.ts
```diff
@@ -307,8 +307,9 @@ export default class TeamHealthService {
     const commitDate = component.latest_commit?.date_time && startOfDay(new Date(component.latest_commit?.date_time))
     const commitSha = component.latest_commit?.sha?.slice(0, 7)
 
-    const baseSha = commitSha
-    const baseDate = component.part_of_monorepo ? latestDevEnvironment?.buildDate || commitDate : commitDate
+    const useLatestDevOverLastCommit = component.part_of_monorepo || latestDevEnvironment?.buildDate >= commitDate
```
andrewrlee, line 310: if monorepo we always want to prefer the dev version as it's likely that the repo commit refers to a service that isn't the one that has been deployed. 

Otherwise use the latest version deployed to dev if it was after or on the same day as the latest main commit. 

There is a little edge case where in very few cases we could still show stale data - if the github discovery runs after the last commit but before the latest version has deployed but this is unlikley and would be fixed by the deployment finishing.

PR #333, opened 2025-06-10
feat/HEAT-537-Alert-Manager-Styling

eve-bramley:
Reintroduced the colours for silenced and active alerts back to the alerts table. Added further styling to prevent table contents from overflowing.


cugorji-justice (approved):
LGTM

PR #345, opened 2025-06-18
Tidying up github request type

andrewrlee:
Example of moving to generically unwrap strapi wrappers in client to continue decoupling rest of the code from strapi


psoleckimoj (approved):
LGTM!

PR #346, opened 2025-06-19
HEAT-679-Alerts-Manager-Table-updates

eve-bramley:
HEAT-679 Made the .label-and-button dropdown selects a consistent size
HEAT-680 Added a new 'Starts at' column
HEAT-681 Moved all 'View' links to a single cell and renamed according to what they link to
HEAT-682 Changed table column ordering and applied consistent column width %

### DIFF
--- a/assets/js/alerts.js
+++ b/assets/js/alerts.js
@@ -165,6 +165,7 @@ function updateURLParams(filters) {
   if (filters.severity) params.set('severity', filters.severity)
   history.replaceState(null, '', `?${params.toString()}`)
 }
+
 // function checks url params for applied filters and builds filter object
 function getFiltersFromURL() {
   const params = new URLSearchParams(location.search)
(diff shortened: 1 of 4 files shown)


assets/scss/alerts.scss
```diff
@@ -2,13 +2,24 @@
 .alert-selectors .label-and-button .govuk-form-group {
     display: flex;
     flex-direction: column;
+    width: 100%;
 }
 
 // For update button alignment with selector
 .alert-selectors .label-and-button .govuk-button {
   margin-top: 15px;
```
psoleckimoj, line 10: You can use `govuk-spacing(3);` for this. Not part of the PR but just for reference 🙂 https://design-system.service.gov.uk/styles/spacing/

psoleckimoj (approved):
LGTM!

assets/js/alerts.js
```diff
@@ -176,42 +177,51 @@ function getFiltersFromURL() {
   }
 }
 
+function formatTimeStamp(dateString) {
+  if (!dateString) return 'N/A'
+  const date = new Date(dateString)
```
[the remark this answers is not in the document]
eve-bramley: Occasionally a past date is provided to the function, so new Date() wouldn't work as the default argument in all cases

assets/scss/alerts.scss
eve-bramley: Couldn't get this to work quite right in this instance, will try and incorporate it going forward

psoleckimoj (approved):
LGTM!

PR #347, opened 2025-06-19
HEAT-668 Address comments on trivy page

andrewrlee:


james-jdgtl (dismissed):
LGTM

server/routes/environments.ts
```diff
@@ -29,9 +29,9 @@ export default function routes({ serviceCatalogueService, redisService }: Servic
     const componentName = getComponentName(req)
     const component = await serviceCatalogueService.getComponent({ componentName })
     const dependencies = (await redisService.getAllDependencies()).getDependencies(componentName)
-    const { environments } = component
-    const prodEnvData = component.environments?.filter(environment => environment.name === 'prod')
-    const alertsSlackChannel = prodEnvData.length === 0 ? '' : prodEnvData[0].alerts_slack_channel
+    const { envs } = component
```
psoleckimoj, line 32: General question, why was this shortened?

server/views/pages/trivyScan.njk
```diff
@@ -65,7 +65,16 @@
           <td>{{ row.severity }}</td>
           <td><a href="{{ row.PrimaryURL }}">{{ row.vulnerabilityID }}</a></td>
           <td>{{ row.pkgName }}</td>
-          <td>{{ row.description }}</td>
+          <td>
```
psoleckimoj, line 68: You should be using the actual components in the templates https://design-system.service.gov.uk/components/details/ so:

```{{ govukDetails({
  summaryText: "Help with nationality",
  text: "We need to know your nationality so we can work out which elections you’re entitled to vote in. If you cannot provide your nationality, you’ll have to send copies of identity documents through the post."
}) }}```

and making sure

```{% from "govuk/components/details/macro.njk" import govukDetails %}```

is at the top

File: server/routes/environments.ts (hunk 1 above)
Comment (s99mj) [reply server/routes/environments.ts:32]: Components have environments and envs both , environment is repeatable component, just  to be clear that i am using envs and from that link its getting environment table.

File: server/data/strapiApiTypes.ts [hunk 3]
```diff
@@ -59,6 +59,8 @@ export type TrivyScanRequest = components['schemas']['TrivyScanRequest']
 export type Env = components['schemas']['Environment']
 export type EnvRequest = components['schemas']['EnvironmentRequest']
 
+export type Env2 = components['schemas']['Component']['envs']['data'][0]['attributes']
```
psoleckimoj, line 62: Needs a better name, `Env2` is not meaningful

server/data/strapiApiClient.ts
```diff
@@ -98,7 +98,7 @@ export default class StrapiApiClient {
   }
 
   async getComponent({ componentName }: { componentName: string }): Promise<SingleResponse<Component>> {
-    const populate = new URLSearchParams({ populate: 'product.team,environments' }).toString()
+    const populate = new URLSearchParams({ populate: 'product.team,envs.trivy_scan' }).toString()
```
psoleckimoj, line 101: Are you sure you want this change? Is anything else using the other environment data? As it's changed what the function returns.

s99mj: Yes that is the only way I will be able to get the latest data for trivy scan for that particular environment

server/routes/components.ts
```diff
@@ -96,20 +96,24 @@ export default function routes({ serviceCatalogueService, redisService, alertsSe
...
     const activeAgencies =
-      environments.length === 0 ? '' : formatActiveAgencies(environments[0].active_agencies as Array<string>)
+      filteredEnvironment.length === 0
+        ? ''
+        : formatActiveAgencies(filteredEnvironment[0].attributes.active_agencies as Array<string>)
     const allowList = new Map()
 
-    if (environments[0].ip_allow_list && environments[0].ip_allow_list_enabled) {
-      const ipAllowListFiles = Object.keys(environments[0].ip_allow_list)
+    if (filteredEnvironment[0].attributes.ip_allow_list && filteredEnvironment[0].attributes.ip_allow_list_enabled) {
```
psoleckimoj, line 108: I would assign `filteredEnvironment[0].attributes` to a new variable as you're using it a lot, making the code look more complicated than it is, same across other areas that are doing this.

psoleckimoj: Maybe `const attributes = filteredEnvironment[0].attributes` 🤷‍♂️

psoleckimoj (approved):
LGTM!

PR #350, opened 2025-06-20
Update Redis modules and other deps

psoleckimoj:
Requiring some help on a type with this one. Followed https://github.com/redis/node-redis?tab=readme-ov-file#blocking-commands to change the `commandOptions({ isolated: true })` hopefully correctly, but the last type error not sure on as the error is as helpful as ever.


server/services/redisService.ts
```diff
@@ -19,7 +19,16 @@ export type VersionDetails = {
 }
 
 export default class RedisService {
-  constructor(private readonly redisClient: RedisClient) {}
+  constructor(
+    private readonly redisClient: RedisClient,
```
andrewrlee, line 23: Do we still need this redisClient?

package.json
```diff
@@ -97,75 +97,75 @@
...
     "@types/connect-flash": "0.0.40",
     "@types/cookie-session": "^2.0.49",
-    "@types/express-session": "^1.18.1",
-    "@types/http-errors": "^2.0.4",
-    "@types/jest": "^29.5.14",
-    "@types/node": "^22.15.18",
+    "@types/express-session": "^1.18.2",
+    "@types/http-errors": "^2.0.5",
+    "@types/jest": "^30.0.0",
+    "@types/node": "^24.0.3",
```
andrewrlee, line 145: did we want this as 24?

PR #355, opened 2025-06-23
count veracode for a team

cugorji-justice:
Count number of Veracode CRITICAL & HIGH vulnerabilities
https://dsdmoj.atlassian.net/browse/HEAT-665

### DIFF
--- a/server/routes/teams.ts
+++ b/server/routes/teams.ts
@@ -37,6 +37,14 @@ export default function routes({ serviceCatalogueService, teamsSummaryCountServi
         serviceCatalogueService,
       )
       logger.info(`getTeamTrivyScanSummary for team '${teamSlug}': ${JSON.stringify(teamTrivyScanSummary, null, 2)}`)
+
+      const teamVeracodeScanSummary = await teamsSummaryCountService.getTeamVeracodeVulnerabilityCounts(
+        products,
+        serviceCatalogueService,
+      )
+      logger.info(
+        `getTeamVeracodeScanSummary for team '${teamSlug}': ${JSON.stringify(teamVeracodeScanSummary, null, 2)}`,
+      )
     } catch (err) {
       logger.error(`Error calling getTeamAlertSummary for team '${teamSlug}':`, err)
     }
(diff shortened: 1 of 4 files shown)


server/services/teamsSummaryCountService.ts
```diff
@@ -218,4 +218,62 @@ export default class TeamsSummaryCountService {
...
+        return productId && productIds.has(productId)
+      })
+
+      const counts = validComponents.reduce(
+        (accumulator, component) => {
+          const summary = component.attributes?.veracode_results_summary as VeracodeResultsSummary | undefined
+          if (!summary?.severity) return accumulator
+
+          summary.severity.forEach(severity => {
+            severity.category.forEach(category => {
```
psoleckimoj, line 246: Just thinking, to save all this you could have a severity type (enum, whatever) and check the severity is in that, then just use the actual severity as the key, then you can do `accumulator[category.severity]` maybe? Then the return object will always be an array of actual severity keys and their values?

Meant to say last time, could just use `total` instead of `accumulator`, easier to write 😄

server/routes/teams.ts
```diff
@@ -37,6 +37,14 @@ export default function routes({ serviceCatalogueService, teamsSummaryCountServi
         serviceCatalogueService,
       )
       logger.info(`getTeamTrivyScanSummary for team '${teamSlug}': ${JSON.stringify(teamTrivyScanSummary, null, 2)}`)
+
+      const teamVeracodeScanSummary = await teamsSummaryCountService.getTeamVeracodeVulnerabilityCounts(
+        products,
+        serviceCatalogueService,
+      )
+      logger.info(
+        `getTeamVeracodeScanSummary for team '${teamSlug}': ${JSON.stringify(teamVeracodeScanSummary, null, 2)}`,
```
psoleckimoj, line 46: Only question is how much data is in the Summary? You don't want to be logging tonnes of data on each request, maybe `logger.debug` at the very least so it's not dumped in prod?

psoleckimoj (approved):
LGTM!

server/routes/teams.ts
cugorji-justice: Not much
 {
    "veryHigh": 0,
     "high": 3,
    "medium": 390,
     "low": 23
}

PR #367, opened 2025-07-01
Moved unwrap in utils and GithubRepoRequest updated

s99mj:


server/utils/utils.ts
```diff
@@ -323,3 +323,48 @@ export const utcTimestampToUtcDate = (str: string) => (str ? formatDate(new Date
 
 export const utcTimestampToUtcDateTime = (str: string) =>
   str ? formatDate(new Date(str), 'dd-MMM-yyyy HH:mm:ss').toUpperCase() : undefined
+
+export function unwrapSingleResponse<T>(response: SingleResponse<T>): T & { id: number } {
+  const data =
+    Array.isArray(response.data) && response.data.length > 0 ? (response.data[0] as DataItem<T>) : response.data
+
+  const unwrapAttributes = (item: Record<string, unknown>): Record<string, unknown> => {
```
andrewrlee, line 331: I'd extract this into a separate function and do some intensive unit tests of it

PR #368, opened 2025-07-01
Starting to add unit test cases for strapi 4 response conversion

andrewrlee:


server/utils/strapi4Utils.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  const result = unwrapAttributes(testSingle.data)
+
+  expect(result).toStrictEqual({
+    id: 1,
+    age: 34,
+    name: 'Test User',
+    nestedSimpleType: {
+      // we should change the recursive algorith to remove this nested data part
```
andrewrlee, line 63: note of something we need to fix going

s99mj (approved):
LGTM

PR #373, opened 2025-07-02
Adding test for array handling

andrewrlee:


server/data/strapiApiClient.ts
```diff
@@ -136,7 +136,7 @@ export default class StrapiApiClient {
     return this.restClient.get({ path, query })
   }
 
-  async getNamespaces(): Promise<Array<Namespace>> {
+  async getNamespaces(): Promise<Unwrapped<Namespace>[]> {
```
andrewrlee, line 139: This makes sense - we should be exposing the new types from the client but ideally we wouldn't expose this `Unwrapped` type everywhere..

server/utils/strapi4Utils.ts
```diff
@@ -36,8 +37,6 @@ export function unwrapSingleResponse<T extends Record<string, unknown>>(response
   return unwrapAttributes<T>(data)
 }
 
-export function unwrapListResponse<T extends Record<string, unknown>>(
-  response: ListResponse<T>,
-): Array<T & { id: number }> {
-  return response.data.map(({ id, attributes }) => ({ ...(unwrapAttributes(attributes) as T), id }))
+export function unwrapListResponse<T extends Record<string, unknown>>(response: ListResponse<T>): Unwrapped<T>[] {
+  return response.data.map(({ id, attributes }) => ({ ...unwrapAttributes<T>(attributes), id }))
```
andrewrlee, line 41: The `unwrapAttributes` being able to understand arrays means we could replace this with:

`unwrapAttributes<T>(response.data) as ...loads of dodgy casting`

but think your `response.data.map` is more easy to understand!

server/utils/utils.ts
```diff
@@ -11,8 +11,8 @@ import type { ServiceCatalogueService } from '../services'
 dayjs.extend(relativeTime.default)
 
 type HasName = { attributes?: { name: string } }
-type HasRepoName = { github_repo: string }
-type HasTeamName = { team_name: string }
+type HasRepoName = { github_repo?: string }
+type HasTeamName = { team_name?: string }
```
andrewrlee, line 15: There was something a bit weird going on here - these `github_repo` and `team_name` are optional on the underlying types but it wasn't complaining before.. now it correctly is so had to make these optional

s99mj (approved):
LGTM

PR #374, opened 2025-07-03
Unwrapped components and replaced environments with envs in getComponent

s99mj:


server/services/teamsSummaryCountService.ts
```diff
@@ -228,39 +225,37 @@ export default class TeamsSummaryCountService {
...
+
+        for (const [severity, count] of severities) {
+          if (VALID_SEVERITIES.includes(severity)) {
+            counts[severity] += count
+          } else {
+            logger.warn(`[getTeamVeracodeVulnerabilityCounts] Unexpected severity: ${severity}`)
+          }
+        }
+      }
+
```
andrewrlee, line 258: Hey @cugorji-justice , we ended up rewriting this a little bit as we had a bug in this area that was difficult to debug. It's equivalent functionality.
 
If you want to have a chat about it then let me know

PR #377, opened 2025-07-07
Fix types issue

andrewrlee:
Remove unnecessary casting


server/data/strapiApiClient.ts
```diff
@@ -126,7 +126,7 @@ export default class StrapiApiClient {
     teamId?: number
     teamSlug?: string
     withEnvironments?: boolean
-  }): Promise<SingleResponse<Team>> {
+  }): Promise<SingleResponse<Team> | ListResponse<Team>> {
```
andrewrlee, line 129: This pattern is manky - 
If it's search by slug it returns a list with one it, if it's search by id it returns a single response.

The service layer always then has to determine which type of response and potentially unwrap the array to get the item out - this should all be handled at this layer so the service layer doesn't need to know that strapi doesn't natively support search by slug

server/data/strapiApiTypes.ts
```diff
@@ -1,12 +1,13 @@
 import { components } from '../@types/strapi-api'
+import { TrivyScanType } from './converters/modelTypes'
 
 export type DataItem<T> = {
-  attributes?: T
+  attributes: T
```
andrewrlee, line 5: these should be mandatory as they are what defines a data item.. etc

server/routes/components.test.ts
```diff
@@ -201,11 +201,17 @@ describe('/components', () => {
...
           expect($('[data-test="description"]').text()).toBe(testComponent.description)
           expect($('[data-test="title"]').text()).toBe(testComponent.title)
-          expect($('[data-test="jira-project-keys"]').text()).toContain(testComponent.jira_project_keys.join(', '))
-          expect($('[data-test="github-write"]').text()).toContain(testComponent.github_project_teams_write.join(', '))
-          expect($('[data-test="github-admin"]').text()).toContain(testComponent.github_project_teams_admin.join(', '))
+          expect($('[data-test="jira-project-keys"]').text()).toContain(
+            (testComponent.jira_project_keys as string[]).join(', '),
+          )
+          expect($('[data-test="github-write"]').text()).toContain(
+            (testComponent.github_project_teams_write as string[]).join(', '),
```
andrewrlee, line 208: these are all unknown on the strapi generated types

server/routes/components.ts
```diff
@@ -104,20 +104,18 @@ export default function routes({ serviceCatalogueService, redisService, alertsSe
       const ipAllowListFiles = Object.keys(envAttributes.ip_allow_list)
       allowList.set('groups', [])
       ipAllowListFiles.forEach(fileName => {
-        // @ts-expect-error Suppress any declaration
         Object.keys(envAttributes.ip_allow_list[fileName]).forEach(item => {
           if (item === 'generic-service') {
-            // @ts-expect-error Suppress any declaration
             const genericService = envAttributes.ip_allow_list[fileName]['generic-service']
             Object.keys(genericService).forEach(ipName => {
```
andrewrlee, line 110: this should all be in the service area

s99mj (dismissed):
LGTM

server/data/strapiApiClient.test.ts
```diff
@@ -163,7 +163,7 @@ describe('strapiApiClient', () => {
     describe('getComponent', () => {
       it('should return a single component', async () => {
         const component = {
-          data: [{ id: 1, attributes: { name: 'component' } }],
+          data: { id: 1, attributes: { name: 'component' } },
```
andrewrlee, line 166: DataItem currently has this as a single value - it could be a list but the types might get super complex. The wrapper handles this nicely so it doesn't really matter

s99mj (approved):
LGTM

PR #378, opened 2025-07-07
Isolating the strapi 4 types

andrewrlee:
Now we should just reference types in the modelTypes.ts file


server/data/strapiApiClient.ts
```diff
@@ -52,11 +50,11 @@ export default class StrapiApiClient {
     }
 
     return this.restClient
-      .get<ListResponse<Product>>({
+      .get<ListResponse<Strapi.Product>>({
```
andrewrlee, line 53: This reflects the general pattern: 

we receive a `ListResponse` of Strapi `Products` and then return an array of model `Products

server/data/strapiApiClient.ts
```diff
@@ -260,95 +256,95 @@ export default class StrapiApiClient {
...
-      .get<SingleResponse<ScheduledJob>>({
+      .get<SingleResponse<Strapi.ScheduledJob>>({
         path: '/v1/scheduled-jobs',
         query: `filters[name][$eq]=${name}`,
       })
-      .then(unwrapSingleResponse<ScheduledJob>)
+      .then(unwrapSingleResponse)
   }
 
   async getTrivyScans(): Promise<TrivyScanType[]> {
```
andrewrlee, line 330: we still need to update this to use unwrapper rather than converter

server/data/strapiApiTypes.ts
```diff
@@ -1,41 +1,12 @@
...
-      total?: number
-    }
-  }
-}
-
-export type SingleResponse<T> = {
-  data: DataItem<T>
-  meta?: Record<string, never>
-}
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
andrewrlee, line 1: All the strapi 4 types.. could rename and tidy up in a follow up

server/routes/components.test.ts
```diff
@@ -4,9 +4,9 @@ import * as cheerio from 'cheerio'
 import { appWithAllRoutes } from './testutils/appSetup'
 import ServiceCatalogueService from '../services/serviceCatalogueService'
 import RedisService from '../services/redisService'
-import { Component, Environment, Product, Unwrapped } from '../data/strapiApiTypes'
```
andrewrlee, line 7: We don't need to refer to `Unwrapped` everywhere

server/data/modelTypes.ts
```diff
@@ -0,0 +1,102 @@
+import type * as Strapi from './strapiApiTypes'
```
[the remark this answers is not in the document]
s99mj: This makes things very easy

PR #397, opened 2025-08-13
Feat/heat 758 pagination on alerts table

amritpal-chahal:
…rs  when data is fetched. Fetch frequency slowed when navigating pages in the alerts table - in the same way as when using the drop-downs. 

Improves UX when you have selected a page number and are scrolling through the alerts table. If the page numbers are not visible on the viewport, when the data would refresh every 5 seconds, the page would scroll down to show the pagination.


psoleckimoj (approved):
LGTM!

PR #399, opened 2025-08-14
Feat/heat 759 alerts table error handling

amritpal-chahal:
Custom error handler for alerts page

- replaced alert pop up with an error message
- functionality for custom errors can be used for other datatables

### DIFF
--- a/assets/js/alerts.js
+++ b/assets/js/alerts.js
@@ -87,6 +87,34 @@ jQuery(function () {
         $(row).addClass('active-alert')
       }
     },
+    ajaxErrorHandler: function (jqXHR, textStatus, errorThrown) {
+      $('#alertsErrorStatus').html(
+        `<div role="region" class="moj-alert moj-alert--error" aria-label="error: Unable to load alerts data. Please try again later" data-module="moj-alert">
+          <div>
+            <svg
+              class="moj-alert__icon"
+              role="presentation"
+              focusable="false"
+              xmlns="http://www.w3.org/2000/svg"
+              viewBox="0 0 30 30"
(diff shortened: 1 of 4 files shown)


assets/js/alerts.js
```diff
@@ -88,6 +88,34 @@ jQuery(function () {
         $(row).addClass('active-alert')
       }
     },
+    ajaxErrorHandler: function (jqXHR, textStatus, errorThrown) {
```
psoleckimoj, line 91: This is fine but as a separate PR just thinking, maybe this could be in a Nunjucks macro which is included in each data tables page njk (can use the proper macro for alerts then) and defaults to hidden, this would then just remove/add a class to show the message? Could also then just change the text? 🤔

assets/js/alerts.js
```diff
@@ -109,6 +137,8 @@ jQuery(function () {
 
   // xhr event is fired when an Ajax request is completed, whether it is successful (data refreshes) or there's an error
   alertsTable.on('xhr', function (_e, settings, json) {
+    $('#alertsErrorStatus').empty()
```
psoleckimoj, line 140: Re my other suggestions, this would be `show()`/`hide()` following that

server/views/pages/alerts.njk
```diff
@@ -148,6 +148,7 @@
   </form>
   
   <p id="lastUpdated"></p>
+  <div id="alertsErrorStatus"></div>
```
psoleckimoj, line 151: Again re my suggestion, could just use the proper `mojAlert` macro here and have it hidden?

PR #406, opened 2025-08-29
slack channel link added to team overview page

amritpal-chahal:


server/views/pages/teamOverview.njk
```diff
@@ -13,6 +13,14 @@
 
 <h1>{% if teamName %}{{ teamName }}{% else %}Team Overview{% endif %}</h1>
 
+{% if slackChannelId and slackChannelName %}
+  <h3 class="team-slack-channel">
+    <image src="/assets/images/slack_favicon_32.png" width="32" height="32"></image> Slack Channel:
+    <a href="slack://channel?team=T02DYEB3A&id={{ slackChannelId }}">#{{ slackChannelName }}</a>
```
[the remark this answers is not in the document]
amritpal-chahal: Good shout, thanks Paul

server/routes/teams.ts
```diff
@@ -65,16 +65,22 @@ export default function routes({
       const channelRecommendations = monitoringChannelService.generateChannelRecommendations(team)
       const channelTree = monitoringChannelService.generateChannelTree(channelRecommendations)
 
-      res.render('pages/teamOverview', {
-        teamName: team.name,
-        formatTeamNameURL: encodeURIComponent(team.name),
-        teamSlug: team.slug,
-        teamAlertSummary,
+      const slackTeamId = 'T02DYEB3A'
```
psoleckimoj, line 68: This really needs to be in config, putting hardcoded values like this in a specific place makes it harder to find for others and also makes it difficult to reuse if required. Also not sure what this ID relates to? We don't have teams as such in Slack and the channel IDs are used below so what actually is this value?

server/routes/teams.ts
```diff
@@ -65,16 +65,22 @@ export default function routes({
...
+
+      const displayTeamOverview = {
```
psoleckimoj, line 70: The const wants naming the same as the name of the attribute you're passing into the template. Is it a team or is it an overview?

server/views/pages/teamOverview.njk
```diff
@@ -4,22 +4,30 @@
...
-{{ breadCrumb(teamName, [{title: "Teams", href: "/teams"}]) }}
+{{ breadCrumb(team.name, [{title: "Teams", href: "/teams"}]) }}
 
-<h1>{% if teamName %}{{ teamName }}{% else %}Team Overview{% endif %}</h1>
+<h1>{% if team.name %}{{ team.name }}{% else %}Team Overview{% endif %}</h1>
 
-{% if teamAlertSummary.total > 0 %}
+{% if team.slackChannelId and team.slackChannelName %}
+  <h3 class="team-slack-channel">
+    <image src="/assets/images/slack_favicon_32.png" width="32" height="32"></image> Slack Channel:
```
psoleckimoj, line 18: Did you mean to use an `` tag? Also it would be self-closing. There is also no `alt` attribute for accessibility or maybe it needs an aria attribute or marked as not required 🤔 It's been a while 😀

psoleckimoj: Actually, if it's not required as part of the back, use it as a `background-image` for the anchor tag and just position it as required.

server/routes/teams.ts
amritpal-chahal: I'll move it to config, I misunderstood when you said const/config so just made a const in teams.ts. Moving it to the config file makes sense, thanks!

The slack team ID is our workspace ID, it is used for deep linking to channels in Slack. It's the same workspace ID used in teams.js for each team's slack channel links. Link to the Slack dev docs, it's under open a channel near the end

https://docs.slack.dev/interactivity/deep-linking/

image

The same slack workspace ID is used in teams.js - I guess I could change this hardcoded value to use it from the config once I move it there too :) 

image

psoleckimoj: Yep all good, you can pass in to the template 🙂

server/routes/teams.ts
psoleckimoj: Nothing wrong with the object per se so happy as is, just the naming of the variable and the attribute you pass in so it is clearer 🙂

server/routes/teams.ts
amritpal-chahal: Moved the ID to config, and called it `slackWorkspaceId` as that's clearer

server/views/pages/teamOverview.njk
amritpal-chahal: Changed Slack icon to a CSS background image as advised as it's simply a visual clue. The text itself is sufficient for accessibility & screen readers etc.

server/config.ts
```diff
@@ -71,4 +71,7 @@ export default {
     },
   },
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
+  slack: {
+    workspaceId: get('SLACK_WORKSPACE_ID', 'T02DYEB3A', requiredInProduction),
```
psoleckimoj, line 75: So this wants to be set in `values-*.yaml`, the value here is the default if not set. If it's not going to change across envs then it should be set in `values.yaml` 👍

server/views/pages/teamOverview.njk
```diff
@@ -157,19 +164,19 @@
...
-              <li class="govuk-hint govuk-body-s">Alternative: <strong>{{ channelRecommendations.suggestedChannels.prodAlt }}</strong> - Shorter prod channel name</li>
+              <li><strong>{{ team.channelRecommendations.suggestedChannels.nonProd }}</strong> - For dev and preprod alerts</li>
+              <li><strong>{{ team.channelRecommendations.suggestedChannels.prod }}</strong> - For production alerts</li>
+              <li class="govuk-hint govuk-body-s">Alternative: <strong>{{ team.channelRecommendations.suggestedChannels.prodAlt }}</strong> - Shorter prod channel name</li>
             </ul>
           </div>
           <div class="govuk-grid-column-one-half">
             <h4 class="govuk-heading-s">Channel Structure</h4>
-            <pre class="govuk-body-s" style="background-color: #f3f2f1; padding: 15px; border-radius: 4px; font-family: monospace;">{% for line in channelTree %}{{ line }}
+            <pre class="govuk-body-s" style="background-color: #f3f2f1; padding: 15px; border-radius: 4px; font-family: monospace;">{% for line in team.channelTree %}{{ line }}
```
psoleckimoj, line 174: Don't use inline styles. This needs turning into a class and put in the css file. It's impossible to override inline styles in the css file, plus you're sending this data every time the page is sent, in the css file it's cached and only the class name is here.

server/views/pages/teams.njk
```diff
@@ -13,7 +13,7 @@
     <thead>
       <tr>
         <th class="govuk-!-width-one-quarter">Name</th>
-        <th class="govuk-!-width-one-quarter"><image src="/assets/images/slack_favicon_32.png" width="32" height="32"></image> Slack (opens app)</th>
+        <th class="govuk-!-width-one-quarter"><img src="/assets/images/slack_favicon_32.png" width="32" height="32" alt="" aria-hidden="true"/> Slack (opens app)</th>
```
psoleckimoj, line 16: As mentioned previously, the image doesn't serve an actual purpose, just put it as a background image on the `th` or create a `span` over the text and use it there.

server/views/pages/teamOverview.njk
```diff
@@ -4,25 +4,32 @@
...
 
 {% block content %}
 
-{{ breadCrumb(teamName, [{title: "Teams", href: "/teams"}]) }}
+{{ breadCrumb(team.name, [{title: "Teams", href: "/teams"}]) }}
 
 <h3>The information on this page relates to products & components in the production environment only</h3>
 
-<h1>{% if teamName %}{{ teamName }}{% else %}Team Overview{% endif %}</h1>
+<h1>{% if team.name %}{{ team.name }}{% else %}Team Overview{% endif %}</h1>
```
andrewrlee, line 16: is it likely you'd ever not have a team.name?

PR #407, opened 2025-09-03
HEAT-829-Change-alert-count-on-team-overview-to-prod-alerts-only

sophiemetcalf:
- Helper function added to teamsSummaryCountServices to filter active alerts for prod environment only. 
- Reflected when using links onb the teams overview page. 
- Notice added to top of team overview page re page only reflecting components & products in prod.


server/services/teamsSummaryCountService.ts
```diff
@@ -145,6 +146,14 @@ export default class TeamsSummaryCountService {
     }
   }
 
+  /**
+   * Helper: Get active alerts for prod environment only
+   */
+  async getActiveAlertsForProd(): Promise<Alert[]> {
```
psoleckimoj, line 152: Is this too specific? It could be more general by passing the env as an arg?

psoleckimoj: Even default to `prod` if nothing passed?

PR #411, opened 2025-09-04
adding unit tests for modelTypes

cugorji-justice:
https://dsdmoj.atlassian.net/browse/HEAT-770


server/data/modelTypes.test.ts
```diff
@@ -0,0 +1,446 @@
+import { Summary, ScanResult, ScanSummary, TrivyScanType, VeracodeResultsSummary } from './modelTypes'
+
+describe('ModelTypes', () => {
+  describe('Type Structure Validation', () => {
+    it('should validate Summary type structure', () => {
```
andrewrlee, line 5: I don't think we need to write tests for types as thats what the compiler tests for us on our behalf

andrewrlee: it might be nice to extract the examples of types into their own test data module where they can be used across tests..

we do something similar [here](https://github.com/ministryofjustice/create-and-vary-a-licence-api/blob/a1b9920bc2428b8d58c6af79d0e5e28ded74bbeb/src/test/kotlin/uk/gov/justice/digital/hmpps/createandvaryalicenceapi/service/TestData.kt#L4) in kotlin.. it can simplify other tests suites if we centralise them?

PR #415, opened 2025-09-10
Moved createStrapiQuery in Utils

s99mj:


server/utils/utils.ts
```diff
@@ -307,14 +307,44 @@ export const utcTimestampToUtcDate = (str: string) => (str ? formatDate(new Date
...
+      }
+      current =
+        typeof current[key] === 'object' && current[key] !== null
+          ? (current[key] as { populate?: Record<string, unknown> }).populate ||
+            (current[key] as Record<string, unknown>)
+          : current
+    })
+  })
+
+  function buildQuery(
```
andrewrlee, line 332: To start with, think it would be good to extract this out of this function.. being nested in the other function makes this much more difficult to read as it implies that you're accessing some of the outer variables (which if you are is a bit of a smell in its own right)

andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #436, opened 2025-09-23
Reusable function for recommended dependency version

cugorji-justice:
https://dsdmoj.atlassian.net/browse/HEAT-876

### DIFF
--- a/server/services/recommendedVersionsService.ts
+++ b/server/services/recommendedVersionsService.ts
@@ -1,6 +1,7 @@
 import logger from '../../logger'
 import config from '../config'
 import ServiceCatalogueService from './serviceCatalogueService'
+import type { Component } from '../data/modelTypes'
 
 export type recommendedVersions = {
   helmDependencies: {
@@ -95,7 +96,7 @@ export default class RecommendedVersionsService {
     return undefined
   }
 
-  private async getComponentByAnyName(variantNames: string[]): Promise<unknown | null> {
+  private async getComponentByAnyName(variantNames: string[]): Promise<Component | null> {
     const [primaryVariant, ...fallbackVariants] = variantNames
 
     // Try primary first (single request)
@@ -165,8 +166,9 @@ export default class RecommendedVersionsService {
         return null
       }
 
-      const versions = (component as unknown as { versions?: Record<string, unknown> }).versions || {}
-      const values = (component as unknown as { values?: Record<string, unknown> }).values || {}
+      type ComponentWithValues = Component & { values?: Record<string, unknown> }
+      const versions = (component.versions || {}) as Record<string, unknown>
+      const values = ((component as ComponentWithValues).values || {}) as Record<string, unknown>
 
       // Preferred keys from versions
       const helmDeps = (versions.helm_dependencies as Record<string, unknown>) || {}


server/services/index.ts
```diff
@@ -29,6 +30,9 @@ export const services = () => {
     strapiApiClientBuilder,
   )
   const monitoringChannelService = new MonitoringChannelService()
+  const recommendedVersionsService = new RecommendedVersionsService()
+  // Inject ServiceCatalogueService so RecommendedVersionsService can fetch hmpps-template-kotlin from Strapi
+  recommendedVersionsService.setServiceCatalogueService(serviceCatalogueService)
```
psoleckimoj, line 35: Can't you just pass this into the constructor when you do `new ....` ?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+    if (raw === null || raw === undefined) return undefined
+    if (typeof raw === 'string' || typeof raw === 'number') return String(raw)
+    if (typeof raw === 'object') {
+      const obj = raw as Record<string, unknown>
+      return (obj.ref as string) || (obj.version as string) || undefined
+    }
+    return undefined
+  }
+
+  private async tryFetchFromStrapi(): Promise<{
```
psoleckimoj, line 99: Function name needs an actual name, not sure what it's supposed to be doing from the current name?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+    private readonly ttlMillis: number = Number(process.env.RECOMMENDED_VERSIONS_TTL_MS) || 6 * 60 * 60 * 1000, // 6h
+  ) {}
+
+  setServiceCatalogueService(service: ServiceCatalogueService): void {
+    this.serviceCatalogueService = service
+  }
+
+  async getRecommendedVersions(): Promise<RecommendedVersions> {
+    const now = Date.now()
+    if (RecommendedVersionsService.cache && RecommendedVersionsService.cache.expiresAt > now) {
```
psoleckimoj, line 44: `this.cache` etc ?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+    // Mark partial if any key is missing
+    if (
+      !result.helm_dependencies.generic_prometheus_alerts ||
+      !result.helm_dependencies.generic_service ||
+      !result.gradle.hmpps_gradle_spring_boot
+    ) {
+      result.metadata.source = result.metadata.source === 'none' ? 'none' : 'partial'
+    }
+
+    RecommendedVersionsService.cache = {
```
psoleckimoj, line 73: Again, why are you not using `this`?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+      value: result,
+      expiresAt: now + this.ttlMillis,
+    }
+
+    logger.info(
```
psoleckimoj, line 78: This is technically only useful for debugging so should be `logger.debug`.

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+      )
+
+      if (helmGenericPrometheusAlerts || helmGenericService || gradleHmppsGradleSpringBoot) {
+        return {
+          helm_generic_prometheus_alerts: helmGenericPrometheusAlerts,
+          helm_generic_service: helmGenericService,
+          gradle_hmpps_gradle_spring_boot: gradleHmppsGradleSpringBoot,
+        }
+      }
+      return null
```
psoleckimoj, line 202: This should be outside the try/catch and also removed from the catch as the final return in the try and catch are both the same.

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+        const vHelm = (values.helm as Record<string, unknown>) || {}
+        const vDeps = (vHelm.dependencies as Record<string, unknown>) || {}
+        helmGenericPrometheusAlerts =
+          helmGenericPrometheusAlerts || this.parseVersionValue(vDeps['generic-prometheus-alerts'])
+        helmGenericService = helmGenericService || this.parseVersionValue(vDeps['generic-service'])
+      }
+
+      const versionsKeys = Object.keys(versions || {})
+      const valuesKeys = Object.keys(values || {})
+      logger.info(
```
psoleckimoj, line 184: Again this is debug info

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+        result.helm_dependencies.generic_prometheus_alerts || 'missing'
+      }, gs=${result.helm_dependencies.generic_service || 'missing'}), gradle(hgsb=${
+        result.gradle.hmpps_gradle_spring_boot || 'missing'
+      })`,
+    )
+
+    return result
+  }
+
+  private parseVersionValue(raw: unknown): string | undefined {
```
psoleckimoj, line 89: Why would you want undefined as a possible value? Why not an empty string?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+ *
+ * It first attempts to read a single versions file (versions.yaml / versions.yml).
+ * If not present, it falls back to:
+ *  - helm_deploy/values.yaml for helm dependency versions
+ *  - gradle/libs.versions.toml for hmpps_gradle_spring_boot version
+ *
+ * Results are cached in-memory with TTL to avoid excessive network calls.
+ */
+export default class RecommendedVersionsService {
+  private static cache: { value: RecommendedVersions; expiresAt: number } | null = null
```
psoleckimoj, line 30: `versions` rather than `value`? `value` singular implies it's not an object

server/services/recommendedVersionsService.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    const svc = new RecommendedVersionsService(ttlMs)
+    const getComponentMock = jest.fn().mockResolvedValue(component)
+    const mockCatalogue = {
+      getComponent: getComponentMock,
+    } as unknown as ServiceCatalogueService
+    svc.setServiceCatalogueService(mockCatalogue)
+    return { svc, mockCatalogue, getComponentMock }
+  }
+
+  it('returns versions from Strapi component (preferred keys)', async () => {
```
psoleckimoj, line 19: These tests aren't passing in an input and checking the output is correct for the given input?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+    try {
+      if (!this.serviceCatalogueService) return null
+      const provided = process.env.HMPPS_TEMPLATE_COMPONENT_NAME || 'hmpps-template-kotlin'
+      const candidates = Array.from(new Set([provided, provided.replace(/_/g, '-'), provided.replace(/-/g, '_')]))
+
+      const tryCandidate = async (idx: number): Promise<unknown | null> => {
+        if (idx >= candidates.length) return null
+        const name = candidates[idx]
+        try {
+          const comp = await this.serviceCatalogueService!.getComponent({ componentName: name })
```
psoleckimoj, line 113: `component` no need to shorten var names like this

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+    helm_generic_service?: string
+    gradle_hmpps_gradle_spring_boot?: string
+  } | null> {
+    try {
+      if (!this.serviceCatalogueService) return null
+      const provided = process.env.HMPPS_TEMPLATE_COMPONENT_NAME || 'hmpps-template-kotlin'
+      const candidates = Array.from(new Set([provided, provided.replace(/_/g, '-'), provided.replace(/-/g, '_')]))
+
+      const tryCandidate = async (idx: number): Promise<unknown | null> => {
+        if (idx >= candidates.length) return null
```
psoleckimoj, line 110: Reliant on an external variable with no checks as to it's existence

server/services/recommendedVersionsService.test.ts
```diff
@@ -0,0 +1,92 @@
...
+describe('RecommendedVersionsService (Strapi only)', () => {
+  beforeEach(() => {
+    ;(RecommendedVersionsService as unknown as { cache: unknown | null }).cache = null
+  })
+
+  function makeSvcWithComponent(component: unknown, ttlMs = -1) {
+    const svc = new RecommendedVersionsService(ttlMs)
+    const getComponentMock = jest.fn().mockResolvedValue(component)
+    const mockCatalogue = {
+      getComponent: getComponentMock,
```
andrewrlee, line 13: **suggestion:** just use `jest.mock` - if you're creating manual mocks for classes you're very likely doing something wrong

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+        const name = candidates[idx]
+        try {
+          const comp = await this.serviceCatalogueService!.getComponent({ componentName: name })
+          logger.info(`[RecommendedVersions] Strapi component match: name=${name}`)
+          return comp
+        } catch (e) {
+          logger.warn(`[RecommendedVersions] Strapi lookup failed for name=${name}: ${String(e)}`)
+          return tryCandidate(idx + 1)
```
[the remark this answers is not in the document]
psoleckimoj: Also you've made a recursive function, it doesn't need to be as far as i can see?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  async getRecommendedVersions(): Promise<RecommendedVersions> {
+    const now = Date.now()
+    if (RecommendedVersionsService.cache && RecommendedVersionsService.cache.expiresAt > now) {
+      return RecommendedVersionsService.cache.value
+    }
+
+    const result: RecommendedVersions = {
+      helm_dependencies: {},
```
andrewrlee, line 49: **suggestion:** if we're in control of this type then we should use standard JS naming conventions and field names are `camelCase`

server/services/recommendedVersionsService.ts
cugorji-justice: Returning undefined carries better semantics than an empty string here.

my understanding of undefined is “not present/not found.” An empty string could mean “explicitly set to blank,” which is a different state.

recommendedVersions type uses optional properties (e.g. generic_prometheus_alerts?: string). Using undefined aligns with that when a value isn’t present, the property can be omitted or left undefined.

I also used it so that I could see a cleaner JSON and logs: JSON.stringify omits properties with undefined values, keeping payloads compact and less ambiguous. An empty string would appear as "", which can be confused for a real value.

So should I still make the changes?

server/services/recommendedVersionsService.test.ts
```diff
@@ -0,0 +1,108 @@
+import RecommendedVersionsService from './recommendedVersionsService'
+import ServiceCatalogueService from './serviceCatalogueService'
+
+jest.mock('./serviceCatalogueService')
+
+// Minimal shape used by RecommendedVersionsService when reading from Strapi
+type StrapiComponentMock = {
+  versions?: {
+    helm_dependencies?: Record<string, unknown>
```
psoleckimoj, line 9: You haven't action @andrewrlee comment about camel-case names.

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,219 @@
...
+ */
+export default class RecommendedVersionsService {
+  private cache: { versions: recommendedVersions; expiresAt: number } | null = null
+
+  private serviceCatalogueService: ServiceCatalogueService
+
+  private readonly ttlMillis: number
+
+  constructor(serviceCatalogueService: ServiceCatalogueService) {
+    this.ttlMillis = config.recommendedVersions.ttlMs
```
psoleckimoj, line 38: Actually can't you just set this at line 35?

server/services/recommendedVersionsService.ts
psoleckimoj: This one isn't resolved.

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,219 @@
...
+          [
+            templateComponentName,
+            templateComponentName.replace(/_/g, '-'),
+            templateComponentName.replace(/-/g, '_'),
+          ].filter(Boolean),
+        ),
+      )
+
+      let component: unknown | null = null
+      for (const variantName of componentNameVariants) {
```
psoleckimoj, line 122: Could refine it further by using `const component = componentNameVariants.find(variantName +> {....` maybe 🤔

server/services/recommendedVersionsService.ts
psoleckimoj: Perfectly valid explanation thanks 👍 🙂

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,219 @@
...
+        )
+        return null
+      }
+
+      const versions = (component as unknown as { versions?: Record<string, unknown> }).versions || {}
+      const values = (component as unknown as { values?: Record<string, unknown> }).values || {}
+
+      // Preferred keys from versions
+      const helmDeps = (versions.helm_dependencies as Record<string, unknown>) || {}
+      let helmGenericPrometheusAlerts = this.parseVersionValue(helmDeps?.generic_prometheus_alerts)
```
psoleckimoj, line 145: Feel free to tell me I'm wrong, but rather than all these possible calls to `parseVersionValue`, could you not just set all the attributes like `helmGenericPrometheusAlerts` to the raw value, then once you've gone through all of your checks, then just reset them via this at the end, so your return would be:

```
return {
          helm_generic_prometheus_alerts: this.parseVersionValue(helmGenericPrometheusAlerts),
          helm_generic_service: this.parseVersionValue(helmGenericService),
          gradle_hmpps_gradle_spring_boot: this.parseVersionValue(gradleHmppsGradleSpringBoot),
        }
```

Maybe?

server/services/recommendedVersionsService.ts
```diff
@@ -0,0 +1,247 @@
...
+    }
+    return undefined
+  }
+
+  private async getComponentByAnyName(variantNames: string[]): Promise<unknown | null> {
+    const [primaryVariant, ...fallbackVariants] = variantNames
+
+    // Try primary first (single request)
+    try {
+      const component = await this.serviceCatalogueService.getComponent({ componentName: primaryVariant })
```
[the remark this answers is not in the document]
cugorji-justice: done, thanks

PR #439, opened 2025-09-25
Added component Archive option

s99mj:


server/routes/componentRequests.ts
```diff
@@ -1,12 +1,137 @@
...
   const router = Router()
 
+  router.get('/choice', async (req, res) => {
+    return res.render('pages/componentRequestFormOption', {
+      title: 'Github Repository Requst Form Options',
+    })
+  })
+
+  router.post('/choice', async (req, res): Promise<void> => {
+    const formData = req.body
```
psoleckimoj, line 17: Not sure this is needed as it takes the same number of characters and doesn't actually do anything?

server/services/componentNameService.ts
```diff
@@ -103,4 +103,12 @@ export default class ComponentNameService {
     const components = componentData.find(repoName => formatMonitorName(repoName.github_repo) === repositoryName)
     return !!components
   }
+
+  async checkComponentArchiveRequestExists(repositoryName: string): Promise<boolean> {
+    const componentData = await this.strapiApiClientFactory('').getGithubRepoRequests()
+    const components = componentData.find(
```
psoleckimoj, line 109: Maybe could use `.some` and then just return here instead of `return !!components` as you don't care about the index?

server/views/pages/index.njk
```diff
@@ -43,7 +43,7 @@
     <div class="govuk-grid-column-one-half">
       <h2>Request Forms</h2>
       <ul>
-        <li><a class="govuk-link--no-visited-state" href="/component-requests/new">Component request form</a> 
+        <li><a class="govuk-link--no-visited-state" href="/component-requests/choice">Component request form</a>
```
psoleckimoj, line 46: Confused by `/choice` as the URL as it's called "Component request form" 🤔 What is it giving the choice of?

server/routes/componentRequests.ts
s99mj: Making it req.body?.* makes rest of the fields as unknown 
Property 'github_repo' does not exist on type 'number'.

server/views/pages/index.njk
s99mj: Choice is Add repo or Archive Repo so named choice

psoleckimoj (approved):
LGTM!

PR #445, opened 2025-10-01
Heat 877 single component page dependencies

sophiemetcalf:


server/routes/components.ts
```diff
@@ -87,6 +92,36 @@ export default function routes({ serviceCatalogueService, redisService, alertsSe
...
+      const recommended = await recommendedVersionsService.getRecommendedVersions()
+      const comparison = compareComponentsDependencies([component], recommended)
+
+      ;(displayComponent as Record<string, unknown>).dependencyComparison = comparison
+
+      const { totalItems, aligned, needsUpgrade, aboveBaseline, missing } = comparison.summary
+      logger.info(
+        `[DependencyComparison] component=${component.name} source=${comparison.recommendedSource} items=${totalItems} aligned=${aligned} needsUpgrade=${needsUpgrade} aboveBaseline=${aboveBaseline} missing=${missing}`,
+      )
+      const nonAligned = comparison.items.filter(i => i.status !== 'aligned')
```
psoleckimoj, line 107: Use meaningful variable names i.e. `item` and not `i`

server/views/pages/component.njk
```diff
@@ -80,16 +80,74 @@
...
+            <th class="govuk-table__header">Dependency</th>
+            <th class="govuk-table__header">Current Version</th>
+            <th class="govuk-table__header">Recommended</th>
+          </tr>
+          </thead>
+          <tbody class="govuk-table__body">
+          {% for item in component.dependencyComparison.items %}
+            {% if item.status == 'needs-attention' %}
+              <tr class="govuk-table__row">
+                <td class="govuk-table__cell" style="white-space: normal; word-break: break-word;">{{ item.key }}</td>
```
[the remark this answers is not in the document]
psoleckimoj: You could even just stick a class on the row and use it to address all `td`s inside it.

server/views/pages/component.njk
```diff
@@ -80,16 +80,74 @@
         </table>
       </div>
     {% endif %}
+
+    {% set upgradeNeeded = false %}
```
psoleckimoj, line 84: You could do this in the route rather then the template then just pass `upgradeNeeded` through in the `res.render` ?

server/views/pages/component.njk
```diff
@@ -53,15 +53,15 @@
               <td colspan="4" class="govuk-table__cell govuk-!-padding-0">
                 <details class="govuk-details" data-module="govuk-details">
```
psoleckimoj, line 54: https://design-system.service.gov.uk/components/summary-list/ ?

server/views/pages/component.njk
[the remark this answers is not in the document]
sophiemetcalf: Removed this so no inline styling

server/views/pages/component.njk
```diff
@@ -22,11 +22,11 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 data-test="detail-page-title">{{ component.name }} {% if component.archived == true %} (Archived){% endif %}</h1>
+    <h1 data-test="detail-page-title">{{ component.name }}</h1>
     {% if component.alerts | length %}
-      <div class="govuk-inset-text govuk-!-margin-bottom-4" data-test="alerts" style="border-left: 5px solid #d4351c; background-color: #fef1f0;">
+      <div id = "component-alert-popup" class="govuk-inset-text govuk-!-margin-bottom-4" data-test="alerts">
```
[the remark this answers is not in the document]
sophiemetcalf: space removed

server/views/pages/component.njk
```diff
@@ -80,16 +80,74 @@
...
+          </tbody>
+        </table>
+      </div>
+    {%  endif %}
+
+    {% if attentionNeeded %}
+      <div class="govuk-inset-text govuk-!-margin-bottom-4" data-test="dependency-attention">
+        <h2 class="govuk-heading-m">Dependencies within recommended range</h2>
+        <p class="govuk-hint">These are aligned to a floating recommendation (e.g., major-only like <code>1.x</code> or minor-only like <code>1.12.x</code>).</p>
+        <table class="govuk-table">
```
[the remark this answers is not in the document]
sophiemetcalf: Removed this!

server/views/pages/component.njk
sophiemetcalf: As above - used table component instead as this was what the original design used (but the njk component).

server/services/dependencyComparison.ts
```diff
@@ -0,0 +1,218 @@
...
+  if (recSegments >= 3) {
+    // Exact version recommendation
+    const cmp = compareVersionTriples(currentParts, recommendedParts)
+    if (cmp < 0) {
+      status = 'needs-upgrade'
+    } else if (cmp === 0) {
+      status = 'aligned'
+    } else {
+      status = 'above-baseline'
+    }
```
psoleckimoj, line 87: I would use early returns, so in here you don't have what `status` was wen you got inside the `if` check. So move line 77 inside the `else`, then change all the `status =` statements to return, the else can also go because of this i.e....

```
if (recSegments >= 3) {
  // Exact version recommendation
  const cmp = compareVersionTriples(currentParts, recommendedParts)
  if (cmp < 0) {
    return 'needs-upgrade'
  } else if (cmp === 0) {
    return 'aligned'
  }
  
  return 'above-baseline'
}

const currentSegs = countVersionSegments(current)
....
```

If you need that debug at the end just sort something inside that `if` statement if that makes sense.

server/services/dependencyComparison.ts
```diff
@@ -0,0 +1,218 @@
...
+  } else {
+    // Floating recommendation logic
+    const currentSegs = countVersionSegments(current)
+
+    if (recSegments === 2) {
+      // Major.minor recommended: compare majors first, then minors
+      const [cMaj, cMin] = [currentParts[0], currentParts[1]]
+      const [rMaj, rMin] = [recommendedParts[0], recommendedParts[1]]
+
+      if (cMaj < rMaj) {
```
psoleckimoj, line 97: TBH, you could do similar here and just return, it would make the code neater. Saves unnecessary `else` statements too as you've already returned.

psoleckimoj: Basically anywhere you do `status = X` you can just return instead.

server/views/pages/component.njk
```diff
@@ -20,76 +22,103 @@
...
+        {%  set additionalAlertsRow = [] %}
+        {%  for alert in component.alerts %}
+          {%  set tableRowContent =  [
+          { text: alert.alertname },
+          { text: alert.environment },
+          { text: alert.summary },
+          { text: alertMessage(alert) }
+          ] %}
+          {% if loop.first %}
+            {%  set _ = firstAlertRow.push(tableRowContent) %}
```
psoleckimoj, line 61: `_` isn't a useful name for a variable?

server/services/dependencyComparison.ts
sophiemetcalf: Removed the status= and used early returns. Removed debug as not really needed now.

server/services/dependencyComparison.ts
sophiemetcalf: As above!

server/services/dependencyComparison.ts
```diff
@@ -0,0 +1,213 @@
...
+      return 'needs-upgrade'
+    }
+    if (cmp === 0) {
+      return 'aligned'
+    }
+    return 'above-baseline'
+  }
+
+  // Floating recommendation logic
+  const currentSegs = countVersionSegments(current)
```
psoleckimoj, line 90: Just spotted this is only needed in the `if` so move it inside 👍

psoleckimoj (approved):
LGTM!

PR #446, opened 2025-10-01
moved inline styling from teamOverview.njk to dedicated scss file, re…

amritpal-chahal:
…moved redundant logic in template and amended team prefix so monitoring channel shows legacy count and correct messages

### DIFF
--- a/assets/scss/teamOverview.scss
+++ b/assets/scss/teamOverview.scss
@@ -1,0 +1,62 @@
+.monitoring-channel {
+  &-warning {
+    background-color: #fef7f7;
+    padding: 20px;
+    border-left: 5px solid #d4351c;
+    margin-bottom: 30px;
+    border-radius: 4px;
+    display: flex;
+    align-items: center;
+    gap: 20px;
+  }
+
+  &-count {
+    background: linear-gradient(135deg, #d4351c 0%, #b91c1c 100%);
+    color: #fff;
+    border-radius: 50%;
+    width: 50px;
+    height: 50px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    font-size: 20px;
+    font-weight: bold;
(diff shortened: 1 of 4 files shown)


assets/scss/teamOverview.scss
```diff
@@ -0,0 +1,62 @@
+.monitoring-channel {
+  &-warning {
+    background-color: #fef7f7;
+    padding: 20px;
+    border-left: 5px solid #d4351c;
```
psoleckimoj, line 5: Could use `$govuk-error-colour`

server/routes/teams.ts
```diff
@@ -67,12 +67,14 @@ export default function routes({
...
-      const hasLegacyChannels = channelRecommendations.recommendations.some(
+      const legacyChannelCount = channelRecommendations.recommendations.filter(
         rec =>
           rec.currentChannels.dev === '#dps_alerts_non_prod' ||
           rec.currentChannels.preprod === '#dps_alerts_non_prod' ||
           rec.currentChannels.prod === '#dps_alerts',
-      )
+      ).length
+
+      const hasLegacyChannels = legacyChannelCount > 0
```
psoleckimoj, line 77: Genuinely, what is the difference here?

`.some()` returns `false` if none are found, otherwise `true`. `filter().length` returns 0 if none are found and > 0 if some are found. This check then checks the count > zero?

Or am I misunderstanding `.some()`?

server/routes/teams.ts
```diff
@@ -86,6 +88,7 @@ export default function routes({
         channelRecommendations,
         channelTree,
         hasLegacyChannels,
```
psoleckimoj, line 90: I've get rid of the `const` and just use `hasLegacyChannels: legacyChannelCount > 0,`

PR #447, opened 2025-10-02
Updated to add pingdom and hmpps application template details in name…

s99mj:
…space page


server/routes/namespaces.ts
```diff
@@ -22,14 +22,74 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
+      pingdom_check: namespace.pingdom_check,
+      hmpps_template: namespace.hmpps_template,
+      rdsInstances: namespace.rds_instance,
+      elasticache_cluster: namespace.elasticache_cluster,
     }
+
     return res.render('pages/namespace', { namespace: displayNamespace })
   })
 
+  router.get('/:namespaceSlug/rds_instance/:rdsInstance', async (req, res) => {
```
psoleckimoj, line 37: We use `-` rather than `_` in URLs, so should be `/rds-instance/`

server/views/pages/namespace.njk
```diff
@@ -11,23 +11,333 @@
...
+        {% else %}
+          <p data-test="no-rds-instances">None</p>
+        {% endif %}
+      </td>
+    </tr>
   </tbody>
 </table>
 
+{% if namespace.rdsInstances | length %}
+  <h1><u> RDS Instances </u></h1>
```
psoleckimoj, line 82: Don't need this underline.

assets/js/namespaces.js
```diff
@@ -12,44 +12,110 @@ jQuery(function () {
...
+            template =>
+              `<a href="/namespaces/${rowData.name}/hmpps_template/${template.tf_label}">${template.tf_label}</a>`,
+          )
+          .join('<br>')
+
+        if (hmppsTemplates) {
+          $(td).html(hmppsTemplates)
+        } else {
+          $(td).text('N/A')
+        }
```
psoleckimoj, line 102: Maybe tidier as a ternary and remove the if/else? ` $(td).html(hmppsTemplates ? hmppsTemplates : 'N/A')`

server/routes/namespaces.ts
```diff
@@ -22,14 +22,74 @@ export default function routes({ serviceCatalogueService }: Services): Router {
   router.get('/:namespaceSlug', async (req, res) => {
     const namespaceSlug = getFormattedName(req, 'namespaceSlug')
     const namespace = await serviceCatalogueService.getNamespace({ namespaceSlug })
-    const rdsInstances = namespace.rds_instance?.map(rdsInstance => rdsInstance)
 
     const displayNamespace = {
       name: namespace.name,
-      rdsInstances,
+      pingdom_check: namespace.pingdom_check,
```
psoleckimoj, line 28: Sorry. missed these, as we're controlling this object, we use camel case for var names/attributes so the local object wants to use `pingdomCheck`, `hmppsTemplate` etc

PR #448, opened 2025-10-08
Feat/heat 482 add GitHub repostory archiving form

s99mj:


server/routes/componentRequests.ts
```diff
@@ -29,9 +152,19 @@ export default function routes({ componentNameService, serviceCatalogueService,
     res.send(componentRequests)
   })
 
-  router.get('/:repo_name', async (req, res) => {
+  router.get('/:repo_name/:request_type', async (req, res) => {
```
psoleckimoj, line 155: `requestType`

psoleckimoj: Actually `repoName` also for consistency as they're actually param names and not in the URL name itself.

server/routes/componentRequests.ts
```diff
@@ -29,9 +152,19 @@ export default function routes({ componentNameService, serviceCatalogueService,
...
     const repoName = req.params.repo_name
-    const componentRequest = await serviceCatalogueService.getGithubRepoRequest({ repoName })
+    const requestType = req.params.request_type
+    const componentRequest = await serviceCatalogueService.getGithubRepoRequest({ repoName }).then(data => {
+      const mappedData = data.map(item => {
```
psoleckimoj, line 159: `return data.filter(item => item.request_type === requestType)[0]`

I think should do the same?

PR #656, opened 2026-04-14
stage HEAT-1222 Add alerts information to product summary boxes

sophiemetcalf:
- Added product and component information to dashboard including alerts, veracode and dependency upgrades

### DIFF
--- a/assets/scss/dashboard.scss
+++ b/assets/scss/dashboard.scss
@@ -1,0 +1,15 @@
+
+#dashboard-alerts-popup {
+  border-left: 5px solid 	#ca3535;
+  background-color: 	#f4d7d7
+}
+
+#dashboard-veracode-popup {
+  border-left: 5px solid 	#54319f;
+  background-color: #ddd6ec
+}
+
+#dashboard-upgrade-popup {
+  border-left: 5px solid 	#1d70b8;
+  background-color: #d2e2f1
+}
--- a/assets/scss/dashboard.scss
+++ b/assets/scss/dashboard.scss
@@ -1,15 +1,15 @@
 
-#dashboard-alerts-popup {
+#dashboard-alerts-panel {
   border-left: 5px solid govuk-colour("red");
   background-color: govuk-colour("red", $variant: "tint-80")
 }
 
-#dashboard-veracode-popup {
+#dashboard-veracode-panel {
   border-left: 5px solid govuk-colour("purple");
   background-color: govuk-colour("purple", $variant: "tint-80")
 }
 
-#dashboard-upgrade-popup {
+#dashboard-upgrade-panel {
   border-left: 5px solid govuk-colour("blue");
   background-color: govuk-colour("blue", $variant: "tint-80")
 }
--- a/assets/scss/dashboard.scss
+++ b/assets/scss/dashboard.scss
@@ -1,15 +1,15 @@
 
 #dashboard-alerts-popup {
-  border-left: 5px solid 	#ca3535;
-  background-color: 	#f4d7d7
+  border-left: 5px solid govuk-colour("red");
+  background-color: govuk-colour("red", $variant: "tint-80")
 }
 
 #dashboard-veracode-popup {
-  border-left: 5px solid 	#54319f;
-  background-color: #ddd6ec
+  border-left: 5px solid govuk-colour("purple");
+  background-color: govuk-colour("purple", $variant: "tint-80")
 }
 
 #dashboard-upgrade-popup {
-  border-left: 5px solid 	#1d70b8;
-  background-color: #d2e2f1
+  border-left: 5px solid govuk-colour("blue");
+  background-color: govuk-colour("blue", $variant: "tint-80")
 }
--- a/assets/scss/dashboard.scss
+++ b/assets/scss/dashboard.scss
@@ -13,3 +13,8 @@
   border-left: 5px solid govuk-colour("blue");
   background-color: govuk-colour("blue", $variant: "tint-80")
 }
+
+.cookie-banner {
+  padding-left: 20px;
+  background-color: #f4f8fb;;
+}
--- a/assets/scss/dashboard.scss
+++ b/assets/scss/dashboard.scss
@@ -16,5 +16,5 @@
 
 .cookie-banner {
   padding-left: 20px;
-  background-color: #f4f8fb;;
+  background-color: govuk-functional-colour(template-background);
 }
--- a/server/routes/dashboard.ts
+++ b/server/routes/dashboard.ts
@@ -4,8 +4,18 @@ import type { Services } from '../services'
 import { cookieService } from '../services/cookieService'
 import { sanitizeCookieInput } from '../services/sanitizeCookieInput'
 import config from '../config'
+import { DisplayAlert } from './products'
+import { formatMonitorName, mapToCanonicalEnv, utcTimestampToUtcDateTime } from '../utils/utils'
+import { Component, Product } from '../data/modelTypes'
+import { countVeracodeHighAndVeryHigh } from '../utils/vulnerabilitySummary'
+import { compareComponentsDependencies } from '../services/dependencyComparison'
+import logger from '../../logger'
 
-export default function routes({ serviceCatalogueService }: Services): Router {
+export default function routes({
+  serviceCatalogueService,
+  alertsService,
+  recommendedVersionsService,
+}: Services): Router {
   const router = Router()
 
   router.use(cookieParser())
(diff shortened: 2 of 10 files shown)


assets/scss/dashboard.scss
```diff
@@ -0,0 +1,15 @@
+
+#dashboard-alerts-popup {
```
theDustRoom: I think for the colours you should use govuk-colour:
https://design-system.service.gov.uk/styles/colour/

server/routes/dashboard.ts
```diff
@@ -81,16 +188,15 @@ export default function routes({ serviceCatalogueService }: Services): Router {
 
   // Route to delete a saved product
   router.post('/saved-products/delete', (req, res) => {
-    const rawIndex = req.body.index
-    const index = Number.parseInt(String(rawIndex), 10)
-    const currentProductsList = cookieService.getFavouritesFromCookie(req.cookies, config.cookieKeys.productNameCookie)
-    if (!Number.isInteger(index) || index < 0 || index >= currentProductsList.length) {
-      return res.redirect(`/dashboard?error=notfound&value=${index}`)
-    }
+    const rawName = req.body.delete.trim()
```
theDustRoom: If req.body.delete is missing or not a string, that throws. It should be sanitised or at least guarded first.

server/routes/dashboard.ts
```diff
@@ -81,16 +188,15 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
+    const currentProductsList = cookieService.getFavouritesFromCookie(
+      req.cookies,
+      config.cookieKeys.productNameCookie.toLowerCase(),
```
theDustRoom: Every other place uses config.cookieKeys.productNameCookie as-is. Lowercasing the cookie key risks reading a different cookie or none at all.

server/routes/dashboard.ts
```diff
@@ -16,16 +26,113 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
     })
+    const componentsNeedingUpdates: string[] = []
+    const bannerPromises = componentArray
+      ?.filter(component => !component.archived)
+      .map(async component => {
+        try {
+          const allAlerts = await alertsService.getAlertsForComponent(component.name)
+          let total = 0
+          if (Array.isArray(allAlerts) && allAlerts.length > 0) {
+            total += 1
```
theDustRoom: That is not a total count, it is effectively a boolean expressed as 0 or 1. Shouldn't this just be a boolean?

server/routes/dashboard.ts
```diff
@@ -16,16 +26,113 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
+            displayProduct.veracodeComponentName.push(component.name)
+          }
+
+          const isKotlin = (component.language || '') === 'Kotlin'
+          const { kotlinOnly } = config.recommendedVersions
+
+          // Dependency comparison for this component
+          if (!kotlinOnly || isKotlin) {
+            try {
+              const recommended = await recommendedVersionsService.getRecommendedVersions()
```
theDustRoom: Couldn't this be fetched once per request, not once per component?

server/routes/dashboard.ts
```diff
@@ -16,16 +26,113 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
       products,
+      product,
+      component: components,
       usersCookiePrefs,
+      filteredProductsObject,
+      alerts: [] as DisplayAlert[],
+      veracodeVulnerabilityCount: 0,
+      veracodeComponentName: [] as string[],
+    }
+    const componentArray: Component[] = []
```
theDustRoom: This is overcomplicated, const componentArray = components.flat() is clearer.

server/routes/dashboard.ts
```diff
@@ -16,16 +26,113 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
     const attemptedProduct = req.query.value as string | undefined
-    const products = await serviceCatalogueService.getProducts({})
+    const products: Product[] = await serviceCatalogueService.getProducts({})
+    const newProductSet = new Set(productsList.map(product => product.trim().toLowerCase()))
+    const filteredProductsObject = products.filter(product => {
+      const normalisedName = (product.name || '').trim().toLowerCase()
+      return newProductSet.has(normalisedName)
+    })
+    const productSlugs = filteredProductsObject.map(product => product.slug)
+    const product: Product[] = await Promise.all(
```
theDustRoom: Variable naming is confusing here as product is an Array, it could be `productsBySlug` or similar

server/routes/dashboard.ts
```diff
@@ -16,16 +26,113 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
+    }
+    const componentArray: Component[] = []
+    components.forEach(component => {
+      component.forEach(item => {
+        componentArray.push(item)
+      })
     })
+    const componentsNeedingUpdates: string[] = []
+    const bannerPromises = componentArray
+      ?.filter(component => !component.archived)
```
theDustRoom: Optional chaining is unnecessary here as componentArray is always defined

server/views/pages/dashboard.njk
```diff
@@ -77,65 +77,156 @@
...
+ {% macro componentsList(componentsInfo)%}
+    {% if dashboard.component | length %}
+      <ul>
+        {% for details in componentsInfo %}
+          <li>
+            <a data-test="component-link" href="/components/{{ details.name }}">
+              {{ details.name }}{% if details.archived == true %} (ARCHIVED){% endif %}
+            </a>
+              {% for name in dashboard.alerts %}
+                {% if name.componentSlug == details.name and name.environment == 'prod' or name.environment == 'production'%}{{ mojBadge({ text: "Active Alert - PROD", classes: "moj-badge--red" }) }}
```
theDustRoom: This doesn't look right - Any alert in production could match even when the component slug does not

server/views/pages/dashboard.njk
```diff
@@ -77,65 +77,156 @@
...
-      key: {
-        text: "Info to be added"
-      },
-      value: {
-        html: "Info to be added"
-      }
-    }
-  ]
-}) }}
+  {% macro alertsInformation(alertsInfo) %}
```
theDustRoom, line 119: You have alertsInfo and then dashboard.alertsInfo here - is this right?

server/views/pages/dashboard.njk
```diff
@@ -77,65 +77,156 @@
...
+              ]
+            }
+        },
+        rows: [
+          {
+             key: {
+               text: 'Description'
+             },
+             value: {
+               html: product.description
```
theDustRoom: If product.description is not guaranteed safe and sanitised before reaching the template, this is a possible XSS risk. If it is plain text, it should be text, not html.

Same principle applies anywhere user or external API data is rendered as HTML

server/views/pages/dashboard.njk
```diff
@@ -32,17 +32,17 @@
     {% endif %}
   </form>
 
-  {% if error == "notfound" %}
+  {% if dashboard.error == "notfound" %}
   {{ mojAlert({
       variant: "error",
       title: attemptedProduct,
```
theDustRoom: You have attemptedProduct and then dashboard.attemptedProduct, is this right?

server/views/pages/dashboard.njk
```diff
@@ -77,65 +77,156 @@
   {% endmacro %}
 
   {% macro allProductsList(products) %}
```
theDustRoom, line 79: You have products here and then dashboard.products below, is this right?

server/views/pages/dashboard.njk
```diff
@@ -77,65 +77,156 @@
...
+                <li>
+                  <a class="govuk-link" href="/components/{{ componentName }}">{{ componentName }}</a></li>
+              {% endfor %}
+            </ul>
+          </div>
+        {% endif %}
+      </div>
+
+      <div class="govuk-grid-column-one-third">
+        {% if upgradeNeeded.length %}
```
theDustRoom: Elsewhere you use the | length filter. In Nunjucks templates, | length is clearer and more idiomatic

theDustRoom (changes requested):
Looks really good, just a few issues

assets/scss/dashboard.scss
sophiemetcalf: They are gov colours! I just used the code, but I have changed them so they are using the govuk-colour function now.

server/routes/dashboard.ts
sophiemetcalf: Removed

server/routes/dashboard.ts
sophiemetcalf: Yes you're right - this was something I thought I was going to use and never did so gave up on it and forgot to remove it! Removed now.

server/routes/dashboard.ts
sophiemetcalf: Yes that makes sense

server/views/pages/dashboard.njk
sophiemetcalf: Ah yes, thats not what I wanted! Have changed to and name.environment == 'prod' or 'production'

sophiemetcalf: That doesn't work either - this looks okay? {% if name.componentSlug == details.name and name.environment == 'prod' or name.componentSlug == details.name and name.environment == 'production'%}

server/views/pages/dashboard.njk
sophiemetcalf: I think so? because the alertsInfo in the macro is just what I named it, but for the for loop I need dashboard.alertsInfo because thats how I'm passing it if that makes sense? Should I rename in argument in the macro to make it less confusing?

server/views/pages/dashboard.njk
sophiemetcalf: Yes its text, I've changed it to text.

server/views/pages/dashboard.njk
sophiemetcalf: No its not, updated!

server/views/pages/dashboard.njk
sophiemetcalf: No, I should use products in the macro and then dashboard.products when I call the macro

server/views/pages/dashboard.njk
sophiemetcalf: Changed!

server/routes/dashboard.ts
sophiemetcalf: removed ?

assets/scss/dashboard.scss
theDustRoom: Yes, this is what I meant.
By using the govuk-colour function, if there are any changes to the GOVUK Frontend you're covered.

server/routes/dashboard.ts
```diff
@@ -16,16 +26,105 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
       usersCookiePrefs,
-    })
+      filteredProductsObject,
+      alerts: [] as DisplayAlert[],
+      veracodeVulnerabilityCount: 0,
+      veracodeComponentName: [] as string[],
+    }
+
+    const componentArray: Component[] = components.flat()
+    const componentsNeedingUpdates: string[] = []
```
theDustRoom: Every promise is returning the same array reference, not a per-component result. So each productResults[i].componentsNeedingUpdates can end up containing the full shared list. You do dedupe later, which masks it a bit, but it is still leaky and confusing.

A cleaner pattern would be to return a local result per component, something like:

```
return {
  activeAlerts,
  needsUpdate: relevantItems.length > 0 ? component.name : null,
  veracodeCount,
  veracodeComponentName: veracodeCount > 0 ? component.name : null,
}
```

With the error state something like:

```
return {
  activeAlerts: [],
  needsUpdate: null,
  veracodeCount: 0,
  veracodeComponentName: null,
}
```

server/routes/dashboard.ts
```diff
@@ -81,16 +180,16 @@ export default function routes({ serviceCatalogueService }: Services): Router {
...
   // Route to delete a saved product
   router.post('/saved-products/delete', (req, res) => {
-    const rawIndex = req.body.index
-    const index = Number.parseInt(String(rawIndex), 10)
-    const currentProductsList = cookieService.getFavouritesFromCookie(req.cookies, config.cookieKeys.productNameCookie)
-    if (!Number.isInteger(index) || index < 0 || index >= currentProductsList.length) {
-      return res.redirect(`/dashboard?error=notfound&value=${index}`)
+    const rawName = req.body.delete
+    if (!rawName || typeof rawName !== 'string') {
+      throw new Error('delete is missing or is not a string')
```
theDustRoom: Bad request data from a form or tampered request shouldn't throw a 500. Better to return a 400, or just redirect back safely.

theDustRoom (changes requested):

server/routes/dashboard.ts
sophiemetcalf: Ah okay. Redirected instead

server/routes/dashboard.ts
theDustRoom: Yes, the same underlying issue remains.

`const componentsNeedingUpdates: string[] = []`

And then in the async map:

```
if (relevantItems.length > 0) {
  componentsNeedingUpdates.push(component.name)
}

return { activeAlerts, componentsNeedingUpdates }
```

Every successful result is still returning the same shared array reference from the outer scope. 

Instead, if it returns something like:

```
return {
  activeAlerts,
  needsUpdate: relevantItems.length > 0 ? component.name : null,
  veracodeCount,
  veracodeComponentName: veracodeCount > 0 ? component.name : null,
}
```

And for the error:

```
return {
  activeAlerts: [],
  needsUpdate: null,
  veracodeCount: 0,
  veracodeComponentName: null,
}
```

You could then aggregate afterwards - something like:

```
const productResults = await Promise.all(bannerPromises)

displayProduct.alerts = productResults.flatMap(result => result.activeAlerts)

displayProduct.veracodeVulnerabilityCount = productResults.reduce(
  (sum, result) => sum + result.veracodeCount,
  0,
)

displayProduct.veracodeComponentName = productResults
  .map(result => result.veracodeComponentName)
  .filter((name): name is string => !!name)

const upgradeNeeded = productResults
  .map(result => result.needsUpdate)
  .filter((name): name is string => !!name)
```

theDustRoom (changes requested):

theDustRoom (approved):
Nice work Sophie

PR #660, opened 2026-04-17
added autcomplete function to input products search box

sophiemetcalf:


theDustRoom (approved):
Nice work again Sophie

PR #747, opened 2026-07-27
Add developer-portal V2 landing page

Thomas-Geraghty:
- Add a new `/developer-portal-v2` route (temporary until we decide to switch out the main landing page
- Add the `landingMasthead` custom component and wire up its SCSS
- Add the `actionLinkList` custom component and wire up its SCSS
- Add the new `landing-page.njk` template with fallback `#` href's for its links until we get the latter pages setup
- Add an alt class for the header to make it match the 2025 govuk restyle

Hope the team doesnt mind but I like to write tests with AAA comments as it makes them way easier to scan IMO.


andrewrlee (approved):
LGTM

PR #753, opened 2026-07-29
Fix single-bundle/ESBuild bugs

Thomas-Geraghty:
- Fix some issues with strict mode complaining about lack of const's making things globals
- Scope the viewServiceAreaVulnerabilities script better by adjusting the id of its related button (would trigger on some other pages otherwise, didn't really do anything but still messy!)


andrewrlee (approved):
LGTM

PR #754, opened 2026-07-29
feat: switch component request alerts to prod/nonprod Slack channels and surface new fields in components UI

mattops:


james-jdgtl (approved):
LGTM!

PR #762, opened 2026-08-05
Update dependencies 2026-08-05

psoleckimoj:


charliesmith-justice (dismissed):
Please can you make sure the docker images you are using are pinned to a sha256 rather than using the tag this is important because the nodejs base images are immutable and also the apk packages they include aren't pinned.

GlennBarber (approved):

andrewrlee (approved):
LGTM

